PR #1, opened 2021-02-05
Update the examples.

smite:
Expands and updates the usage examples in the notebook.

### DIFF
--- a/docs/usage.ipynb
+++ b/docs/usage.ipynb
@@ -109,9 +109,9 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "QB1: ───Y^0.5────X───ZZ───────S^-1─────────────────────────────\n",
+      "QB1: ───Y^0.5────X───ZZ────────S─────────────────────────────\n",
       "                     │\n",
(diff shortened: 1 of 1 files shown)


docs/usage.ipynb
```diff
@@ -159,16 +200,84 @@
     }
    },
    "source": [
-    "-----\n",
-    "See `examples/qasm_demo.py` for more examples\n"
+    "Note that the above output vector represents the state before the measurement due to the optimized circuit, not the original one, which would have a plus sign instead of a minus. `IQMDevice.simplify_circuit` has eliminated a `ZPowGate` which has no effect on the measurement.\n",
```
olliahonen, line 203:
```suggestion
    "Note that the above output vector represents the state before the measurement in the optimized circuit, not the original one, which would have a plus sign instead of a minus. `IQMDevice.simplify_circuit` has eliminated a `ZPowGate` which has no effect on the measurement.\n",
```

olliahonen: If I got what you mean there, I think it reads more clearly like this.

olliahonen (approved):

PR #11, opened 2021-04-13
Add workflow for tox test

### LINKED ISSUE #5: Automated daily tests against latest Cirq version
Since (potentially breaking) changes to Cirq would not automatically test against the code in `cirq-on-iqm`, we should setup automated tests (using github actions?) in `cirq-on-iqm` running against the latest version of Cirq to raise an alert whenever there's a breaking change so that it can be detected early on and necessary migrations can be done. 

cc @balopat

Spellstaker:
Fixes #5

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,7 +3,7 @@ name: Continuous Integration
 on:
   pull_request:
   schedule:
-    - corn: "0 0 * * *"
+    - cron: "0 0 * * *"
 
 jobs:
   build:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,4 +20,3 @@ jobs:
         run: pip install -e ".[dev]"
       - name: Run Tox
         run: tox -e py
-        
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,8 +15,9 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@v2
         with:
-          python-version: '3.8'
+          python-version: '3.9'
       - name: Install dependencies
         run: pip install -e ".[dev]"
       - name: Run Tox
         run: tox -e py
+        
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,19 @@
+name: Continuous Integration 
+
+on: [ pull_request ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.8'
+      - name: Install Tox
+        run: pip install tox
+      - name: Run Tox
+        run: tox -e py
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,6 +17,6 @@ jobs:
         with:
           python-version: '3.8'
       - name: Install dependencies
-        run: pip install -e ".[dev,docs]"
+        run: pip install -e ".[dev]"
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,19 @@
+name: Continuous Integration 
+
+on: [ pull_request ]
```
tanujkhattar, line 3: This should run on a fixed schedule (daily?) so that we can detect any breaking changes in Cirq. The idea is to ensure that if Cirq updates which breaks cirq-on-iqm, we should be able to detect that early.

tanujkhattar (changes requested):

.github/workflows/ci.yml
Spellstaker: Added schedule to run tests every midnight (UTC).

.github/workflows/ci.yml
```diff
@@ -0,0 +1,22 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.8'
+      - name: Install dependencies
+        run: pip install -e ".[dev,docs]"
```
manevala, line 20: Is this step needed?

Spellstaker: I am using the dev packages here to install the specific version of tox that we are using.
Removing docs.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,22 @@
...
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.8'
```
smite, line 18:
```suggestion
          python-version: '3.9'
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,22 @@
...
+      - name: Install dependencies
+        run: pip install -e ".[dev]"
+      - name: Run Tox
+        run: tox -e py
```
smite, line 22:
```suggestion
        run: tox -e py

```

smite: To get rid of that missing newline warning

.github/workflows/ci.yml
Spellstaker: Updated.

.github/workflows/ci.yml
Spellstaker: Done.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,22 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.9'
+      - name: Install dependencies
+        run: pip install -e ".[dev]"
+      - name: Run Tox
+        run: tox -e py
```
tanujkhattar, line 22: What happens if the tests fail? Do the maintainers / codeowners get notified somehow?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,22 @@
...
+  schedule:
+    - cron: "0 0 * * *"
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
```
tanujkhattar, line 14: I think it would be worth testing against the latest released version as well as the latest pre-released version corresponding to the master branch of the repo.

manevala:
Good points in the comments, I'll add those to our internal backlog :) I'll merge this for now so that we can get at least some kind of CI/CD pipeline running.

PR #14, opened 2021-04-21
Reset version to 0.1

smite:
* The package name has changed to `cirq-iqm`, so we also reset the version history.
* CHANGELOG items should from now on have links to pull requests that implement the feature or fix the bug, done using a `sphinx.ext.extlinks` RST role.


manevala (approved):
👍

dc914337 (approved):

PR #21, opened 2021-06-23
Remove superfluous gate classes, simplify gate decompositions

### LINKED ISSUE #3: [Discuss] IsingGate design and implementation
The [comment in `IsingGate` class](https://github.com/iqm-finland/cirq-on-iqm/blob/870b445240167ca6aa165ac2a54c5bc2e90501fd/src/cirq_iqm/iqm_gates.py#L38) says that it's needed because `cirq.ops.ZZPowGate`  contains decomposition logic which iqm wants to avoid. 

Since the decomposition logic of a gate is mainly called when the code dealing with the gate (eg: Simulator) doesn't understand what the gate is, it's not clear to me how is the default decomposition logic of `ZZPowGate` in Cirq causing an issue for IQM? 

If the problem is that because the hardware natively supports `IsingGate` and therefore we want to use it as a primitive in cirq circuits, I think it would be better if we implement it by inheriting from `cirq.ops.ZZPowGate` and passing the global phase as an init parameter. A similar approach is used in Cirq to implement [`Mølmer–Sørensen Gate`](https://github.com/quantumlib/Cirq/blob/a61e51b53612735e93b3bb8a7605030c499cd6c7/cirq/ion/ion_gates.py#L27) by extending `cirq.XXPowGate`.

Another approach could be to use `ZZPowGate` in the cirq circuits (for optimisation and simulation) and convert to an appropriate representation (of `IsingGate`) -- which the hardware understands -- during the serialisation step when the circuit has to be sent to the hardware. For Eg: See [Serialiser of IonQ](https://github.com/quantumlib/Cirq/blob/7b2abd80ceefe69a24d6708b223d56bd96bb8553/cirq/ionq/serializer.py#L194) that does something similar. 

cc @balopat

smite:
* Removes the `IsingGate` and `XYGate` classes in `iqm_gates`, and replaces them with functions that return the corresponding Cirq native class instances.
* Simplifies the gate decompositions, which now also preserve global phase.
* Removes the `IQMEjectZ` class, which is now unnecessary.
* Some cleanup.

Closes #3.


src/cirq_iqm/adonis.py
```diff
@@ -69,45 +72,42 @@ class Adonis(idev.IQMDevice):
...
-            # the ISwap family is implemented using the XY interaction
-            s = -0.5 * op.gate.exponent
-            return [
-                ig.XYGate(exponent=s).on(*op.qubits)
-            ]
         if isinstance(op.gate, ops.CZPowGate):
-            # decompose CZPowGate using IsingGate
-            s = -0.5 * op.gate.exponent
-            L = ops.ZPowGate(exponent=-s, global_shift=-0.5)
+            # decompose CZPowGate using ZZPowGate
```
manevala, line 79: Minor: This decomposition is shared between Adonis and Valkmusa, would it make sense to create separate shared functions for the decompositions used in both?

src/cirq_iqm/valkmusa.py
```diff
@@ -14,10 +14,23 @@
 """
 IQM's Valkmusa quantum architecture.
 """
+from math import pi as PI
+
 from cirq import ops
 
 import cirq_iqm.iqm_device as idev
-import cirq_iqm.iqm_gates as ig
+#import cirq_iqm.iqm_gates as ig
```
manevala, line 22: This import can probably be removed? :)

src/cirq_iqm/adonis.py
```diff
@@ -16,20 +16,23 @@
...
 # common gates used in gate decompositions
 CZ = ops.CZPowGate()
-Lx = ops.XPowGate(exponent=0.5, global_shift=-0.5)
-Lxi = ops.XPowGate(exponent=-0.5, global_shift=-0.5)
-Ly = ops.YPowGate(exponent=0.5, global_shift=-0.5)
-Lyi = ops.YPowGate(exponent=-0.5, global_shift=-0.5)
+Lx = ops.rx(PI_2)
+Lxi = ops.rx(-PI_2)
+Ly = ops.ry(PI_2)
+Lyi = ops.ry(-PI_2)
```
manevala, line 36: Are these new definitions compatible with the current `iqm_operation_mapping.py` (which is not in this PR but is already in master)? Or will a circuit consisting of these gates evne be assumed to be executable without some further processing?

src/cirq_iqm/adonis.py
smite: That's currently the only shared part. Maybe later if there will be more of them.

src/cirq_iqm/valkmusa.py
smite: Done.

src/cirq_iqm/adonis.py
smite: rx and ry are convenience functions that return XPowGate and YPowGate instances, but use a more common parameterization.

manevala (approved):

PR #22, opened 2021-06-24
Update the project setup. Generate API docs using sphinx.ext.autosummary.

smite:
* Adds a `pyproject.toml` file.
* Removes the PyScaffold dependency.
* Upgrades Sphinx to version 4.0.2.
* Generates API docs using recursive `sphinx.ext.autosummary` instead of `sphinx-automodapi`.
* Adds a `sphinx.ext.intersphinx` mapping.
* Adds `tox` environments for building docs and distribution packages.
* Misc. cleanups.
* Fixes the CI script, which previously executed a non-existing Tox environment.


docs/conf.py
```diff
@@ -71,8 +72,28 @@
...
 
-# directory for the automodapi generated doc files
-automodapi_toctreedirnm = 'api/api'
+# -- Autodoc ------------------------------------------------------------
+
+# member ordering in autodoc output (default: 'alphabetical')
+autodoc_member_order = 'bysource'
+
+# mock some imports when building the docs
+#autodoc_mock_imports = ["django"]
```
olliahonen, line 81:
```suggestion
```

docs/conf.py
```diff
@@ -71,8 +72,28 @@
...
 
 
-# directory for the automodapi generated doc files
-automodapi_toctreedirnm = 'api/api'
+# -- Autodoc ------------------------------------------------------------
+
+# member ordering in autodoc output (default: 'alphabetical')
+autodoc_member_order = 'bysource'
+
+# mock some imports when building the docs
```
olliahonen, line 80:
```suggestion
```

docs/conf.py
```diff
@@ -71,8 +72,28 @@
 show_authors = True
 
 
-# directory for the automodapi generated doc files
-automodapi_toctreedirnm = 'api/api'
+# -- Autodoc ------------------------------------------------------------
+
+# member ordering in autodoc output (default: 'alphabetical')
+autodoc_member_order = 'bysource'
+
```
olliahonen, line 79:
```suggestion
```

README.rst
```diff
@@ -79,24 +79,24 @@ Clone the repository and install it in editable mode with all the extras:
...
+   $ pip install -e ".[dev,docs,testing]"
 
 
 Build and view the docs:
 
 .. code-block:: bash
 
-   python setup.py docs
-   firefox build/sphinx/html/index.html
+   $ tox -e docs
```
hay-k, line 91: Following these steps I get the following error

> Warning, treated as error:
> html_static_path entry '_static' does not exist
> ERROR: InvocationError for command 'cirq-on-iqm\.tox\docs\Scripts\sphinx-build.EXE' -b html docs 'cirq-on-iqm/build/sphinx/html' -W (exited with code 2)
> ________________________ summary ____________________________
> ERROR:   docs: commands failed

src/cirq_iqm/iqm_device.py
```diff
@@ -310,6 +310,7 @@ def optimization_at(
             index: int,
             op: cirq.Operation,
     ) -> Optional[cirq.PointOptimizationSummary]:
+        """TODO the parent class at Cirq has a broken docstring here, which we have to override."""
```
olliahonen, line 313: Can we add a link to the corresponding Cirq issue for fixing the broken docstring?

README.rst
smite: Fixed.

src/cirq_iqm/iqm_device.py
smite: I haven't reported it. There are some broken docstrings in Cirq, but they only cause us trouble because of the "treat warnings as errors" flag we use when building docs. I think Cirq currently uses a custom API docs generation system from TensorFlow, so it may not be a problem for them.

smite: Another solution would be to not inherit any docstrings from Cirq parent classes, but IMO that would make the docs less useful to the user.

docs/conf.py
smite: Removed.

src/cirq_iqm/iqm_device.py
olliahonen: Ok. If we write here that some Cirq docstrings are broken I think we should report it as well. Or if it isn't worth reporting then maybe we don't write that they're broken?

smite: I reported it, https://github.com/quantumlib/Cirq/issues/4276

manevala (approved):
👍

src/cirq_iqm/iqm_device.py
olliahonen:
```suggestion
        """TODO the parent class at Cirq has a broken docstring here, which we have to override."""
        """https://github.com/quantumlib/Cirq/issues/4276"""
```

olliahonen (approved):

PR #27, opened 2021-07-01
Improve naming and descriptions

olliahonen:
* Server or quantum computer instead of backend
* Other improvements to grammar and descriptions
* No changes to functionality

Closes COMP-124

### DIFF
--- a/examples/demo_iqm_engine.py
+++ b/examples/demo_iqm_engine.py
@@ -24,7 +24,7 @@ from cirq_iqm.iqm_device import IQMQubit
 from cirq_iqm.iqm_remote import IQMSampler
 
 
-def demo_adonis_backend():
+def demo_adonis_run():
     """
     Run a circuit on the Adonis quantum computer.
     """
@@ -43,7 +43,7 @@ def demo_adonis_backend():
     with open(os.environ['IQM_SETTINGS_PATH'], 'r') as f:
         sampler = IQMSampler(os.environ['IQM_SERVER_URL'], f.read(), qubit_mapping)
 
-    # This will send the circuit to the backend to be run, and return a cirq.study.Result
+    # This will send the circuit to the server to be run, and return a cirq.study.Result
     # containing the measurement results.
     results = sampler.run(circuit_adonis, repetitions=1000)
 
@@ -57,4 +57,4 @@ def demo_adonis_backend():
 
 
 if __name__ == '__main__':
-    demo_adonis_backend()
+    demo_adonis_run()
--- a/examples/demo_iqm_engine.py
+++ b/examples/demo_iqm_engine.py
@@ -26,7 +26,7 @@ from cirq_iqm.iqm_remote import IQMSampler
 
 def demo_adonis_run():
     """
-    Run a circuit on the Adonis quantum computer.
+    Run a circuit on an Adonis quantum computer.
     """
     qubit = IQMQubit(1)
     circuit = cirq.Circuit(
--- a/src/cirq_iqm/iqm_client.py
+++ b/src/cirq_iqm/iqm_client.py
@@ -125,7 +125,7 @@ class IQMClient:
             qubit_mapping: list[SingleQubitMappingDTO],
             shots: int = 1
     ) -> UUID:
-        """Submits a quantum circuit to be executed on the quantum computer.
+        """Submits a quantum circuit to be executed on a quantum computer.
 
         Args:
             circuit: circuit to be executed
--- a/src/cirq_iqm/iqm_client.py
+++ b/src/cirq_iqm/iqm_client.py
@@ -74,7 +74,7 @@ class SingleQubitMappingDTO(BaseModel):
 
 
 class RunRequestDTO(BaseModel):
-    """Request for IQM quantum computer to execute a circuit.
+    """Request for an IQM quantum computer to execute a circuit.
     """
     qubit_mapping: list[SingleQubitMappingDTO]
     circuit: CircuitDTO
@@ -109,7 +109,7 @@ class RunResult(BaseModel):
 
 
 class IQMClient:
-    """Provides access to an IQM quantum computer.
+    """Provides access to IQM quantum computers.
 
     Args:
         url: Endpoint for accessing the server. Has to start with http or https.
--- a/src/cirq_iqm/iqm_client.py
+++ b/src/cirq_iqm/iqm_client.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """
-Client for calling the IQM computation backend.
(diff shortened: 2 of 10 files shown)


src/cirq_iqm/iqm_remote.py
```diff
@@ -66,12 +66,12 @@ def run_sweep(
             params: cirq.Sweepable,
             repetitions: int = 1,
     ) -> list[cirq.Result]:
-        """Samples from the given quantum circuit.
+        """Samples from, i.e. executes, the given quantum circuit.
```
manevala, line 69: "Samples from the given quantum circuit." comes directly from the the parent class, not sure if changing it brings more clarity.

src/cirq_iqm/iqm_remote.py
```diff
@@ -66,12 +66,12 @@ def run_sweep(
...
 
         Nontrivial sweeping is not yet supported, ``params`` should be ``None``.
 
         Args:
-            program: The circuit to sample from.
+            program: The circuit to execute.
```
manevala, line 74: Same as above. "The circuit to sample from" comes from the parent class.

manevala (approved):

examples/demo_iqm_engine.py
```diff
@@ -24,7 +24,7 @@
 from cirq_iqm.iqm_remote import IQMSampler
 
 
-def demo_adonis_backend():
+def demo_adonis_run():
     """
     Run a circuit on the Adonis quantum computer.
```
smite, line 29:
```suggestion
    Run a circuit on an Adonis quantum computer.
```

src/cirq_iqm/iqm_client.py
```diff
@@ -74,7 +74,7 @@ class SingleQubitMappingDTO(BaseModel):
 
 
 class RunRequestDTO(BaseModel):
-    """Request for IQM backend to execute a circuit.
+    """Request for IQM quantum computer to execute a circuit.
```
smite, line 77:
```suggestion
    """Request for IQM quantum computing service to execute a circuit.
```

src/cirq_iqm/iqm_client.py
```diff
@@ -108,11 +108,11 @@ def from_dict(inp: dict[str, Union[str, dict]]) -> RunResult:
         return RunResult(status=RunStatus(input_copy.pop("status")), **input_copy)
 
 
-class IQMBackendClient:
-    """Provides access to a remote IQM quantum device.
+class IQMClient:
+    """Provides access to an IQM quantum computer.
```
smite, line 112:
```suggestion
    """Provides access to quantum computers through the IQM computation control service.
```

src/cirq_iqm/iqm_client.py
```diff
@@ -108,11 +108,11 @@ def from_dict(inp: dict[str, Union[str, dict]]) -> RunResult:
...
 
     Args:
-        url: Endpoint for accessing the device. Has to start with http or https.
+        url: Endpoint for accessing the server. Has to start with http or https.
```
smite, line 115:
```suggestion
        url: Endpoint for accessing the service. Has to start with http or https.
```

src/cirq_iqm/iqm_client.py
```diff
@@ -125,7 +125,7 @@ def submit_circuit(
             qubit_mapping: list[SingleQubitMappingDTO],
             shots: int = 1
     ) -> UUID:
-        """Submits a quantum circuit to be executed on the backend.
+        """Submits a quantum circuit to be executed on the quantum computer.
```
smite, line 128:
```suggestion
        """Submits a quantum circuit to be executed on a quantum computer.
```

src/cirq_iqm/iqm_remote.py
```diff
@@ -49,15 +49,15 @@ def _serialize_iqm(circuit: cirq.Circuit) -> CircuitDTO:
...
 
 class IQMSampler(cirq.work.Sampler):
-    """Circuit sampler for evaluating quantum circuits on IQM quantum devices.
+    """Circuit sampler for executing quantum circuits on IQM quantum computers.
 
     Args:
-        url: Endpoint for accessing the device. Has to start with http or https.
-        settings: Settings for the quantum hardware.
-        qubit_mapping: A dict that maps qubit logical names to physical qubit names
+        url: Endpoint for accessing the server interface. Has to start with http or https.
```
smite, line 55:
```suggestion
        url: Endpoint for accessing the computation service. Has to start with http or https.
```

src/cirq_iqm/iqm_remote.py
smite: Let's keep it as it was. The user won't call this method directly anyway.

src/cirq_iqm/iqm_remote.py
```diff
@@ -93,17 +93,17 @@ def _send_circuit(
             circuit: cirq.Circuit,
             repetitions: int = 1,
     ) -> cirq.study.Result:
-        """Sends the circuit to be executed by the remote IQM device.
+        """Sends the circuit to be executed on the IQM quantum computer.
```
smite, line 96:
```suggestion
        """Sends the circuit to be executed by the IQM computation service.
```

src/cirq_iqm/iqm_remote.py
olliahonen: AFAIK this is supposed to be one of the functions the user may call directly. But anyway, no point in copying the superclass docstring verbatim, I'll just delete it here. IDEs and our docs generation are able to pick up the docstring from the superclass.

olliahonen: Docstring removed here.

src/cirq_iqm/iqm_remote.py
olliahonen: Docstring removed here.

src/cirq_iqm/iqm_remote.py
olliahonen: Turns out the docstring was not verbatim copy of the superclass one.

olliahonen: Reworded.

src/cirq_iqm/iqm_remote.py
olliahonen: I'll remove the controversial part. It reads just as well like that :) Underscore method, after all.

src/cirq_iqm/iqm_client.py
olliahonen: I see your point here but it's not so easy to keep it consistent across all the modules... It seems that "server/server interface" and "quantum computer" is the least confusing way to explain what we have here. Additional concepts such as a service, however real, only seem to complicate the documentation.

Also, we should be proud of our main product at IQM, the quantum computers :)

PR #30, opened 2021-07-07
Add action for publishing releases to PyPI.

manevala:


.github/workflows/release.yml
```diff
@@ -0,0 +1,30 @@
...
+        with:
+          python-version: '3.9'
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -e ".[dev,cicd]"
+      - name: Build and publish
+        run: |
+          tox -e build
+          python -m twine upload dist/*
```
hay-k, line 27:
```suggestion
          python -m twine check dist/*
          python -m twine upload dist/*
```

setup.cfg
```diff
@@ -64,6 +64,9 @@ testing =
     pytest-pylint == 0.18.0
     jsons==1.4.2
     mockito==1.2.2
+cicd =
+    twine == 3.3.0
+    wheel == 0.36.2
```
smite, line 69: Isn't `wheel` included in the Python image? Is there some reason to pin it?

manevala: The virtual environment used for the action is this one: https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md which afaik does not contain `wheel`. No reason to pin it to that specific version though, should be fixed in my newest commit.

.github/workflows/release.yml
smite:
```suggestion
          python -m twine upload --verbose dist/*
```

smite: Could be useful.

smite (approved):

PR #33, opened 2021-07-14
Publish documentation on GitHub Pages with an Action

### LINKED ISSUE #32: Add action for compiling and publishing docs to GitHub Pages
Add action for compiling and publishing docs to GitHub Pages for example on push to *main*

Running
```
tox -e docs
```
and publishing output in *gh-pages* using [JamesIves/github-pages-deploy-action](https://github.com/marketplace/actions/deploy-to-github-pages) or similar should work.

nikosavola:
### This PR:
* Adds GitHub Action for compiling and deploying documentation to `gh-pages`
* Adds link to documentation in README

`tox -e docs` is used for compiling docs and `touch build/sphinx/html/.nojekyll` is executed afterwards to allow having a `_static` folder where, e.g., the CSS-files are stored, in GitHub Pages.

Closes #32


.github/workflows/docs.yml
```diff
@@ -0,0 +1,31 @@
...
+          sudo apt-get install graphviz
+      - name: Compile docs
+        run: |
+          tox -e docs
+          touch build/sphinx/html/.nojekyll  # allow underscores in URL path
+      - name: Deploy docs
+        uses: JamesIves/github-pages-deploy-action@4.1.4
+        with:
+          branch: gh-pages
+          folder: build/sphinx/html
```
hay-k, line 31: Let's move this job to `release.yml`. This way
- publishing docs will be part of releasing a new version and we will have a single workflow for this purpose
- publishing docs will happen only when a new version is released instead of each push to the `main` branch, so that the latest pip installable verison and the online docs are synchronous.

Also please rename the job from `docs` to `publish_docs`, the step `Compile docs` to `Build docs`, and the step `Deploy docs` to `Publish to gh-pages`.

README.rst
```diff
@@ -20,6 +20,7 @@ Currently Cirq on IQM can
 * run the circuit on an IQM quantum computer
 
 See the `Jupyter Notebook <https://jupyter.org/>`_ with examples: ``examples/usage.ipynb``.
+Documentation is available at `iqm-finland.github.io/cirq-on-iqm/ <https://iqm-finland.github.io/cirq-on-iqm/>`_.
```
hay-k, line 23: Also fix the link to documentation in `setup.cfg`

nikosavola:
I also noticed that in `push_to_pypi` one could change the 

```yaml
    steps:
      - uses: actions/checkout@v2
      - name: Fetch all history for all tags and branches
        run: git fetch --prune --unshallow
```

to

```yaml
    steps:
      - uses: actions/checkout@v2
        with:
          fetch-depth: 0  
```

Would this be better?

hay-k:
> I also noticed that in `push_to_pypi` one could change the
> 
> ```yaml
>     steps:
>       - uses: actions/checkout@v2
>       - name: Fetch all history for all tags and branches
>         run: git fetch --prune --unshallow
> ```
> 
> to
> 
> ```yaml
>     steps:
>       - uses: actions/checkout@v2
>         with:
>           fetch-depth: 0  
> ```
> 
> Would this be better?

Let's leave it as is. The current explicit implementation is more readable.

hay-k:
I removed the documentation link from the README for now. There is an upcoming documentation update which will add the link and release a new version. Until then let's not have a non working URL in the README.

hay-k (approved):

PR #35, opened 2021-08-05
IQMDevice overhaul

### LINKED ISSUE #2: Remove the hack of iterating a fixed number of times in simplify_circuit
Fix the current [hack of using a fixed number of iterations](https://github.com/iqm-finland/cirq-on-iqm/blob/870b445240167ca6aa165ac2a54c5bc2e90501fd/src/cirq_iqm/iqm_device.py#L240) in `simplify_circuit` by directly comparing old and new circuit to check whether any optimisation was applied. 

The change would look something like this:
```python
while True:
    old_circuit = circuit.copy()
    MergeOneParameterGroupGates().optimize_circuit(circuit)
    optimizers.merge_single_qubit_gates_into_phased_x_z(circuit)
    # all z rotations are pushed past the first two-qubit gate following them
    IQMEjectZ(eject_parameterized=True).optimize_circuit(circuit)
    optimizers.DropEmptyMoments().optimize_circuit(circuit)
    # Circuit has reached a fixed point. 
    if circuit == old_circuit:
        break
```

smite:
* Removes `IQMDevice.map_circuit`.
* Adds `IQMDevice.decompose_circuit` and `IQMDevice.route_circuit`.
* `IQMSampler` checks that circuit connectivity fits in the device connectivity.
* Removes the hack in `IQMDevice.simplify_circuit`.
* Fixes a bug in `DropRZBeforeMeasurement`.
* Examples fixed.
* More tests.

Closes #2.
COMP-107


CHANGELOG.rst
```diff
@@ -3,17 +3,41 @@ Changelog
...
+
+* ``IQMDevice`` updated. `#35 <https://github.com/iqm-finland/cirq-on-iqm/pull/35>`_
+
+  * ``IQMDevice.map_circuit`` removed.
+  * ``IQMDevice.decompose_circuit`` and ``IQMDevice.route_circuit`` methods added.
+  * ``IQMDevice.simplify_circuit`` now checks if it has hit a fixed point after each iteration.
+  * ``IQMSampler`` checks that the circuit respects the device connectivity.
+
+* Device qubit handling is simplified. ``IQMSampler`` can generate a trivial qubit mapping automatically.
+  `#34 <https://github.com/iqm-finland/cirq-on-iqm/pull/34>`_
```
hay-k, line 20:
```suggestion
* Device qubit handling is simplified. `#34 <https://github.com/iqm-finland/cirq-on-iqm/pull/34>`_
  * ``IQMSampler`` can generate a trivial qubit mapping automatically.
  * The class ``IQMQubit`` removed.
```

src/cirq_iqm/iqm_remote.py
```diff
@@ -122,6 +122,14 @@ def run_sweep(
         elif program.device != self._device:
             raise ValueError('The devices of the given circuit and of the sampler are not the same.')
 
+        # check that the circuit connectivity fits in the device connectivity
+        # apply qubit_mapping
+        temp = {cirq.NamedQubit(k): cirq.NamedQubit(v) for k, v in self._qubit_mapping.items()}
```
olliahonen, line 127:
```suggestion
        qubit_map = {cirq.NamedQubit(k): cirq.NamedQubit(v) for k, v in self._qubit_mapping.items()}
```

src/cirq_iqm/iqm_remote.py
```diff
@@ -122,6 +122,14 @@ def run_sweep(
...
+        mapped = program.transform_qubits(temp)
```
olliahonen, line 128:
```suggestion
        qubit_map = program.transform_qubits(temp)
```

src/cirq_iqm/iqm_remote.py
olliahonen: `temp` is not very descriptive

tests/test_adonis.py
```diff
@@ -281,25 +327,33 @@ def qubit_mapping(self):
...
+            cirq.CZ(*qubits[0:2]),
+            cirq.CZ(*qubits[1:3]),
+        )
+        new = adonis.route_circuit(circuit)
 
-        assert mapped_circuit.device == adonis
-        assert mapped_circuit.all_qubits() == frozenset(cirq.NamedQubit.range(1, 3, prefix=adonis.QUBIT_NAME_PREFIX))
+        assert len(new.all_qubits()) == 3
+        assert new.all_qubits() <= set(adonis.qubits)
+        #assert len(new) == len(circuit)  # TODO at the moment the routing algo may add unnecessary SWAPs
```
olliahonen, line 346:
```suggestion
```

olliahonen: Seems like a feature of the routing algorithm. If we change the algo then we'll update the tests anyway. Or maybe I misunderstood the point of this test case?

examples/demo_common.py
```diff
@@ -16,28 +16,48 @@
 Demonstrates transforming a quantum circuit into the native gateset and connectivity of a given
 IQM device, optimizing it, and then executing it on a simulator.
 """
+from __future__ import annotations
+
 import cirq
 import numpy as np
```
hay-k, line 22:
```suggestion
from typing import Optional

import cirq
import numpy as np

from cirq_iqm import IQMDevice
```

olliahonen (changes requested):

tests/test_adonis.py
```diff
@@ -281,25 +327,33 @@ def qubit_mapping(self):
...
+        circuit = cirq.Circuit([cirq.X(q) for q in qubits])
+        with pytest.raises(ValueError, match='Number of logical qubits is greater than number of physical qubits'):
+            adonis.route_circuit(circuit)
 
-    def test_indexed_names(self, adonis):
-        qubits = cirq.NamedQubit.range(1, 3, prefix='qubit_')
-        circuit = cirq.Circuit(cirq.measure(*qubits))
-        mapped_circuit = adonis.map_circuit(circuit)
+    def test_routing_without_SWAPs(self, adonis, qubits):
+        # circuit graph can be embedded in the Adonis connectivity graph
```
hay-k, line 337:
```suggestion
        # circuit graph can be embedded in the Adonis connectivity graph without SWAPs
```

tests/test_adonis.py
```diff
@@ -281,25 +327,33 @@ def qubit_mapping(self):
...
+        circuit = cirq.Circuit(
+            cirq.CZ(*qubits[0:2]),
+            cirq.CZ(*qubits[1:3]),
+        )
+        new = adonis.route_circuit(circuit)
 
-        assert mapped_circuit.device == adonis
-        assert mapped_circuit.all_qubits() == frozenset(cirq.NamedQubit.range(1, 3, prefix=adonis.QUBIT_NAME_PREFIX))
+        assert len(new.all_qubits()) == 3
+        assert new.all_qubits() <= set(adonis.qubits)
```
hay-k, line 345:
```suggestion
        assert new.all_qubits() <= adonis.qubit_set()
```

tests/test_adonis.py
hay-k, line 346:
```suggestion
        # assert len(new) == len(circuit)  # TODO at the moment the routing algo may add unnecessary SWAPs
```

examples/demo_common.py
```diff
@@ -16,28 +16,48 @@
...
-def demo(device, circuit_original, do_measure, *, use_qsim=False):
+def demo(
+    device: IQMDevice,
+    circuit: cirq.Circuit,
+    do_measure: bool,
+    *,
+    qubit_mapping: Optional[dict[str, str]] = None,
+    use_qsim: bool = False
+):
     """Tranform the given circuit to a form the given device accepts, then simulate it.
```
hay-k, line 35:
```suggestion
    """Transform the given circuit to a form the given device accepts, then simulate it.
```

tests/test_adonis.py
smite: That's right, it's a feature of the algorithm. The todo just reminds the reader that the particular routing case here is solvable without adding any SWAPs, and the algorithm could perform better.

src/cirq_iqm/iqm_device.py
```diff
@@ -214,63 +183,105 @@ def decompose_operation(self, operation: cirq.Operation) -> cirq.OP_TREE:
...
+
+        Returns:
+            routed circuit
+
+        Raises:
+            ValueError: routing is impossible
+        """
+        device_graph = nx.Graph(tuple(map(self.get_qubit, edge)) for edge in self.CONNECTIVITY)
+        swap_network = route_circuit(circuit, device_graph, algo_name='greedy')
+        return swap_network.circuit
```
hay-k, line 203: The value of the `device` property of the returned circuit is `cirq.UNCONSTRAINED_DEVICE`. Should we update the circuit before returning, so that it is tied to the correct device (i.e. `self`)? 

Updating the device of the circuit results into a call to the `validate_circuit()` method of the new device. So theoretically this might even result into a `ValueError` (if for some reason routing exits normally but does the routing incorrectly). Maybe it is a good idea to catch this exception and re-throw with an updated informative message?

src/cirq_iqm/iqm_device.py
```diff
@@ -214,63 +183,105 @@ def decompose_operation(self, operation: cirq.Operation) -> cirq.OP_TREE:
...
-        DropRZBeforeMeasurement().optimize_circuit(circuit)
-        optimizers.DropEmptyMoments().optimize_circuit(circuit)
-        DecomposeGatesFinal().optimize_circuit(circuit)
+            if c == before:
+                # the optimization hit a fixed point
+                break
+
+        DropRZBeforeMeasurement().optimize_circuit(c)
+        optimizers.DropEmptyMoments().optimize_circuit(c)
+        #DecomposeGatesFinal().optimize_circuit(c)
```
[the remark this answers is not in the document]
smite: Fixed.

tests/test_adonis.py
olliahonen: Okay.

olliahonen (approved):

src/cirq_iqm/iqm_device.py
smite: We can't, since the routing step may add SWAPs which aren't native to the device, which the user needs to decompose using `decompose_circuit`.

PR #43, opened 2021-08-31
Move IQM client code to separate repo.

manevala:


setup.cfg
```diff
@@ -36,8 +36,8 @@ install_requires =
     numpy
     cirq ~= 0.11
     ply  # required by cirq.contrib.qasm_import
-    requests>=2.25.1,<3.0
     pydantic>=1.8.1,<2.0
+    iqm-client>=1.0.0
```
hay-k, line 40:
```suggestion
    iqm-client>=1.0
```

src/cirq_iqm/iqm_sampler.py
```diff
@@ -23,12 +23,13 @@
...
 from cirq.study import resolver
+from iqm_client.iqm_client import Circuit, IQMClient, SingleQubitMapping
+
 from cirq_iqm import IQMDevice
-from cirq_iqm.iqm_client import CircuitDTO, IQMClient, SingleQubitMappingDTO
 from cirq_iqm.iqm_operation_mapping import map_operation
 
 
-def serialize_circuit(circuit: cirq.Circuit) -> CircuitDTO:
+def serialize_circuit(circuit: cirq.Circuit) -> Circuit:
```
hay-k, line 32: Was there a particular reason for renaming all FooDTO classes to just Foo? I think now they are confusing, especially this one, since it is easy to confuse with CIrq's Circuit class.

manevala: I didn't think it added anything and our naming conventions in other repos are different so it was also inconsistent. The third reason is that the pulse compiler uses the objects in data processing, not just in serialization logig. Cirq on IQM uses the Circ Circuit as cirq.Circuit so if there is a risk of confusion, our Circuit could be used as iqm_client.Circuit or imported as IQMCircuit.

tests/test_iqm_gates.py
```diff
@@ -13,14 +13,13 @@
 # limitations under the License.
 """Test IQM parameterizations of certain two-qubit gate families.
 """
-# pylint: disable=no-self-use
-from cirq import ops
 import numpy as np
 import pytest
+# pylint: disable=no-self-use
```
hay-k, line 18: Looks like this line is not necessary. Not sure why it was there historically.

setup.cfg
```diff
@@ -36,8 +36,8 @@ install_requires =
     numpy
     cirq ~= 0.11
     ply  # required by cirq.contrib.qasm_import
-    requests>=2.25.1,<3.0
-    pydantic>=1.8.1,<2.0
+    pydantic >= 1.8.1, < 2.0
```
smite, line 39: Can we remove the pydantic dependency as well?

src/cirq_iqm/iqm_sampler.py
```diff
@@ -23,12 +23,14 @@
 import numpy as np
 from cirq import study
 from cirq.study import resolver
+from iqm_client import iqm_client
+from iqm_client.iqm_client import IQMClient, SingleQubitMapping
```
smite, line 27: It would be more consistent to either import all classes directly, or use `iqm_client.ClassXXX` for all of them.

smite: The latter seems to be the recommended way, more informative to the reader.

manevala: This was discussed and I switched to this style because having a `Circuit` (as opposed to `iqm_client.Circuit`) was deemed confusing (because there's also `cirq.Circuit`). I'm personally fine with using just `Circuit` or `iqm_client.Circuit` but I wouldn't want to preface all imports with `iqm_client` when there is no risk of confusion with Cirq types.

setup.cfg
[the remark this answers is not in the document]
manevala: Good find, fixed.

smite (approved):

PR #46, opened 2021-09-15
PhasedRx OpenQASM import and export, tests.

smite:
* The extended OpenQASM 2.0 parser now imports `U` gates of the form `U(a, b, -b)` as `PhasedXPowGate`.
* OpenQASM import and export tests.

COMP-178


src/cirq_iqm/extended_qasm_parser.py
```diff
@@ -15,20 +15,39 @@
...
     Returns:
         parsed circuit
     """
     parser = QasmParser()
+
+    def convert_U(args: tuple) -> cirq.Gate:
+        """Maps the OpenQASM builtin one-qubit gate ``U`` to :class:`cirq.QasmUGate`,
+        or :class:`cirq.PhasedXPowGate` if possible.
+        """
+        theta, phi, lambd = args
```
olliahonen, line 38:
```suggestion
        theta, phi, lambda = args
```

src/cirq_iqm/extended_qasm_parser.py
```diff
@@ -15,20 +15,39 @@
...
+        scale = np.pi
+        if phi == -lambd:
```
olliahonen, line 40:
```suggestion
        if phi == -lambda:
```

src/cirq_iqm/extended_qasm_parser.py
olliahonen: Minor

src/cirq_iqm/extended_qasm_parser.py
olliahonen: Minor

src/cirq_iqm/extended_qasm_parser.py
manevala: `lambda` is a reserved word in Python and so it can't be used as a variable name. I think `lambd` is fine.

src/cirq_iqm/extended_qasm_parser.py
manevala: Same as above, can't use `lambda` as a variable.

src/cirq_iqm/extended_qasm_parser.py
olliahonen: Ah yes 🙂 In that case I suggest to use the same abbreviation as `QasmUGate` uses: `lmda`

src/cirq_iqm/extended_qasm_parser.py
manevala, line 40: Does it make sense to allow some epsilon's worth of tolerance here? The OpenQASM file might contain gates where the parameters have had some floating point operations done so we might have lost actual equality.

olliahonen (commented):
.

src/cirq_iqm/extended_qasm_parser.py
smite: From a physics perspective yes, if `phi` and `-lambd` are almost equal then the gate is almost PhasedRx. But when an actual PhasedRx is exported, `phi` and `-lambd` should be the exact same float with the same string presentation.

manevala: Ok, that makes sense too 🤔 So, I'm fine with either this or near equality :)

manevala (approved):

olliahonen (approved):

olliahonen (approved):

PR #52, opened 2021-09-30
Package installation to notebook as per Cirq notebook guidelines

### LINKED ISSUE #8: Adopt Cirq Notebook guidelines for tutorial notebooks
Cirq has [guidelines for notebooks](https://github.com/quantumlib/Cirq/blob/master/docs/dev/notebooks.md) and it would be nice to align with it. 

An immediate action item would be to add a header cell that includes commands for packages installation (`pip install` statements) so that the notebook can be run independently on Google Colab. 

cc @balopat

manevala:
See https://github.com/quantumlib/Cirq/blob/master/docs/dev/notebooks.md

Fixes #8

### DIFF
--- a/examples/usage.ipynb
+++ b/examples/usage.ipynb
@@ -6,9 +6,14 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "import cirq\n",
-    "from cirq_iqm import Adonis, circuit_from_qasm\n",
-    "from cirq_iqm.optimizers import simplify_circuit"
+    "try:\n",
+    "    import cirq\n",
+    "    from cirq_iqm import Adonis, circuit_from_qasm\n",
+    "    from cirq_iqm.optimizers import simplify_circuit\n",
+    "except ImportError:\n",
+    "    print('Installing missing dependencies...')\n",
+    "    !pip install --quiet cirq cirq_iqm\n",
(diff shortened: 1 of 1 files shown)


examples/usage.ipynb
```diff
@@ -6,9 +6,14 @@
...
-    "from cirq_iqm import Adonis, circuit_from_qasm\n",
-    "from cirq_iqm.optimizers import simplify_circuit"
+    "try:\n",
+    "    import cirq\n",
+    "    from cirq_iqm import Adonis, circuit_from_qasm\n",
+    "    from cirq_iqm.optimizers import simplify_circuit\n",
+    "except ImportError:\n",
+    "    print('Installing missing dependencies...')\n",
+    "    !pip install --quiet cirq cirq_iqm\n",
+    "    print('Installation ready')"
```
LeonWubben, line 16: If cirq is not installed and thus is installed in the except clause, it won't import Adonis, circuit_from_qasm and simplify_circuit from cirq_iqm. The try/excepts should either be separated for both cirq and cirq_iqm, or you'd need to import everything again after the `!pip install`

manevala: True! Should be hopefully fixed now :)

LeonWubben (approved):

PR #60, opened 2021-11-16
Add favicon to docs

manevala:


docs/conf.py
```diff
@@ -108,6 +108,7 @@
 # Output file base name for HTML help builder.
 htmlhelp_basename = 'cirq_iqm-doc'
 
+html_favicon = 'images/favicon.ico'
```
smite, line 111:
```suggestion
html_favicon = '_static/images/favicon.ico'
```

smite: This is where our other projects store static images, for consistency it'd be good to do it here as well.

smite: `html_static_path` should be uncommented as well.

smite: I checked the docs, https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-html_favicon, and they seem to hint that the "correct" place for favicons would be `_static/`. Who knows :)

olliahonen (approved):

PR #62, opened 2021-11-22
Add api token

manevala:


hay-k:
When I use wrong credentials, I get the following error: `requests.exceptions.HTTPError: 401 Client Error: Unauthorized for url: http://localhost:9002/circuit/run`. Does not seem very user friendly. In the logs of IQM server I can see `Could not log in user with username myname`, which is a bit more understandable. I don't know where the error gets lost, in IQM client or cirq-on-iqm, but Can we do so that the error message from the server is delivered to the end user?

hay-k:
When I use wrong credentials, I get the following error: `requests.exceptions.HTTPError: 401 Client Error: Unauthorized for url: http://localhost:9002/circuit/run`. Does not seem very user friendly. In the logs of IQM server I can see `Could not log in user with username myname`, which is a bit more understandable. I don't know where the error gets lost, in IQM client or cirq-on-iqm, but Can we do so that the error message from the server is delivered to the end user?

hay-k (approved):

PR #66, opened 2022-01-26
Checkout last tagged commit in publish workflow

ollityrkko:
In order to get clean release number, the docs must be built on a tagged commit.
Add checkout step to `publish_docs` job to get to the last tagged commit.
Tags are available only if `actions/checkout@v2` fetches the full history.
For that to happen it needs parameter `fetch-depth: 0`.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -41,6 +41,8 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
       - name: Setup Python
         uses: actions/setup-python@v2
         with:
@@ -50,6 +52,10 @@ jobs:
           python -m pip install --upgrade pip
           pip install -e ".[dev, docs]"
           sudo apt-get install graphviz
+      - name: Checkout latest tagged commit
+        run: |
+          git checkout $(git tag -l | tail -1)
+          tox -e build
       - name: Build docs
         run: |
           tox -e docs
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -55,7 +55,6 @@ jobs:
       - name: Checkout latest tagged commit
         run: |
           git checkout $(git tag -l | tail -1)
-          tox -e build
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -50,6 +52,10 @@ jobs:
           python -m pip install --upgrade pip
           pip install -e ".[dev, docs]"
           sudo apt-get install graphviz
+      - name: Checkout latest tagged commit
+        run: |
+          git checkout $(git tag -l | tail -1)
+          tox -e build
```
hay-k, line 58: I suppose this step is needed only in the case when the publish job is triggered manually, right? If it is triggered automatically, then it is triggered because a new tag/release was created and the last commit is tagged, so `fetch-depth: 0` is enough.

hay-k: I am thinking, do we really need this? What are the use cases of triggering the job manually. Maybe it need not be truncated to the last tagged commit?

ollityrkko: I don't know about the need of it, but `workflow_dispatch` is there, so manual triggering is possible, and the checkout step is there to handle such usage.

hay-k: Yeah, I guess there is no harm. The only situation I can think of is the situation that the automatic run fails and someone needs to rerun it manually.

ollityrkko: The checkout step is only needed in `publish_docs` job.  Code builds are done on the latest commit, tagged or not.

hay-k: BTW, why is `tox -e- build` part of this step? The actual building of the docs is done in the next step with `tox -e docs`

hay-k: Hmm, I am not sure how it helps actually. I just did a quick testing locally and it does not seem to help at all. We are using `setuptools-scm` for guessing the version from git metadata. Maybe this tools has some options to handle this situation more robustly?

hay-k (approved):

PR #72, opened 2022-05-02
Apollo 20-qubit architecture support

freetonik:
Support for 20-qubit IQM Apollo architecture.

The Apollo device differs from Adonis only in two aspects: number of qubits and the connectivity of qubits. The native gates are the same in both Apollo and Adonis, so decomposition logic is also the same.

### DIFF
--- a/examples/demo_apollo.py
+++ b/examples/demo_apollo.py
@@ -1,4 +1,4 @@
-# Copyright 2020–2021 Cirq on IQM developers
+# Copyright 2020–2022 Cirq on IQM developers
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/examples/demo_apollo.py
+++ b/examples/demo_apollo.py
@@ -1,0 +1,58 @@
+# Copyright 2020–2021 Cirq on IQM developers
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+Demonstrates importing a quantum circuit from an OpenQASM 2.0 file, transforming it into
+the Apollo native gateset and connectivity, and then executing it on a simulator.
+"""
+import cirq
+from demo_common import demo
+
+from cirq_iqm.devices import Apollo
+from cirq_iqm.extended_qasm_parser import circuit_from_qasm
+
+
+def demo_apollo(do_measure=False, use_qsim=False):
+    """Run the demo using the Apollo architecture."""
+
+    print('\nApollo demo\n===========\n')
+    device = Apollo()
+    qasm_program = """
+        OPENQASM 2.0;
+        include "qelib1.inc";
+        qreg q[6];
+        creg meas[3];
+        U(0.2, 0.5, 1.7) q[1];
+        h q[0];
+        h q[2];
+        h q[4];
+        h q[5];
+        cx q[2], q[1];
+        cx q[3], q[5];
+    """
+    if do_measure:
+        qasm_program += '\nmeasure q -> meas;'
+
+    circuit = circuit_from_qasm(qasm_program)
+
+    # add some more gates
+    q0 = cirq.NamedQubit('q_0')
+    q5 = cirq.NamedQubit('q_5')
+    circuit.insert(len(circuit) - 1, cirq.CXPowGate(exponent=0.723)(q5, q0))
+
+    qubit_mapping = {'q_0': 'QB1', 'q_1': 'QB2', 'q_2': 'QB3', 'q_3': 'QB4', 'q_4': 'QB5', 'q_5': 'QB6'}
+    demo(device, circuit, do_measure, use_qsim=use_qsim, qubit_mapping=qubit_mapping)
+
+
+if __name__ == '__main__':
+    demo_apollo()
--- a/examples/demo_apollo.py
+++ b/examples/demo_apollo.py
@@ -37,8 +37,8 @@ def demo_apollo(do_measure=False, use_qsim=False):
         h q[2];
         h q[4];
         h q[5];
-        cx q[2], q[1];
-        cx q[3], q[5];
+        cx q[0], q[1];
+        cx q[3], q[4];
     """
     if do_measure:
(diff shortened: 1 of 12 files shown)


examples/demo_apollo.py
```diff
@@ -0,0 +1,58 @@
...
+        include "qelib1.inc";
+        qreg q[6];
+        creg meas[3];
+        U(0.2, 0.5, 1.7) q[1];
+        h q[0];
+        h q[2];
+        h q[4];
+        h q[5];
+        cx q[2], q[1];
+        cx q[3], q[5];
```
smite, line 41: Are these qubits coupled?

examples/usage.ipynb
```diff
@@ -260,6 +258,137 @@
...
+    "        /  |   /  |  /  |\n",
+    "     QB18  QB15  QB11  QB7\n",
+    "        |  /  |  /  |  /\n",
+    "        QB14  QB10  QB6\n",
+    "        /  |  /  |  /\n",
+    "     QB13  QB9    QB5\n",
+    "        |  /  |  /  |\n",
+    "        QB8   QB4   QB2\n",
+    "          |  /  |  /\n",
+    "          QB3   QB1\n",
```
smite, line 286: I would suggest printing this from the Apollo class docstring.

examples/demo_apollo.py
freetonik: No, and neither are the ones on line 51 (`circuit.insert(len(circuit) - 1, cirq.CXPowGate(exponent=0.723)(q5, q0))`). Fixed in next commit, thanks!

examples/usage.ipynb
freetonik: You mean instead of markdown do something like this:

```
apollo = Apollo()
print(apollo.__doc__)
```

?

If so, then probably do the same for Adonis (same notebook, few cells above)?

src/cirq_iqm/devices/apollo.py
```diff
@@ -0,0 +1,149 @@
...
+        ops.XPowGate,
+        ops.YPowGate,
+        ops.MeasurementGate
+    )
+
+    NATIVE_GATE_INSTANCES = (
+        ops.CZPowGate(),
+    )
+
+    def operation_decomposer(self, op: cirq.Operation) -> Optional[list[cirq.Operation]]:
```
hay-k, line 98: Should we move this function to the parent class instead of copying in various device implementations?

freetonik: I guess. But then we'll have to choose between:

1. `IQMDevice` stays as generic as now. `operation_decomposer` moves there, and becomes parameterized to allow for decomposition of different architectures. 
3. `IQMDevice` stays as generic as now. We introduce subclasses for compatible architectures (i.e. single subclass for Adonis&Apollo, since native gates are the same). `operation_decomposer` moves to that subclass.
4. `IQMDevice` becomes less generic and implements `operation_decomposer`, which somehow works for current Adonis&Apollo architecture, and is flexible for future different architectures.

Also, is it possible that two chips have the same native gates, but their decompositions still differ in certain details (e.g. global shift)?

hay-k: I think
1. is not great, since will quickly become cumbersome.
2. is something I had in mind as well, but probably does not make sense to implement now, maybe in the future?
3. is actually what I was thinking: a default operation decomposer in `IQMDevice`, and children can overwrite it as needed.
4. another though I had was to reconsider if we still need this functionality. Why is Cirq's default decomposition enough?

Decomposition of individual gates into the same native gate set should always be the same regardless of device. Circuit optimization routines could be different for different devices and end up in different circuits depending on device, but individual gate decomposition is always the same.

freetonik: @hay-k regarding 4: seems like Cirq default decomposition can handle `ISwapPowGate`. But that's not the case for `CZPowGate`, `ZZPowGate` and `ZPowGate` in our tests. Removing them from `operation_decomposer` break the tests.

So I propose for now to proceed with 3, i.e. moving `operation_decomposer` to `IQMDevice` while removing `ISwapPowGate` decomposition. Then investigate the possibility of getting rid of `operation_decomposer` in a separate issue.

@smite, we might need your feedback regarding this.

examples/usage.ipynb
freetonik: Assuming you meant exactly that: see latest change.

src/cirq_iqm/devices/apollo.py
freetonik: Proceeded with 3: `operation_decomposer` is now implemented in `IQMDevice`, and children devices can override it (see Valkmusa device as an example of such overriding). 

The possibility of getting rid of `operation_decomposer` altogether will be discussed in a separate issue.

examples/usage.ipynb
```diff
@@ -260,6 +251,122 @@
...
+    "circuit.append(cirq.YPowGate(exponent=1.1)(e))\n",
+    "circuit.append(cirq.PhasedXPowGate(phase_exponent=0.3, exponent=0.5)(f))\n",
+    "print(circuit)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Insert non-native gates and decompose them into native ones by calling `decompose_circuit`"
```
hay-k, line 313:
```suggestion
    "Insert non-native gates and decompose the circuit so that it contains native gates only"
```

examples/usage.ipynb
```diff
@@ -260,6 +251,122 @@
...
+    "circuit.append(cirq.HPowGate(exponent=-0.4)(f))\n",
+    "circuit = apollo.decompose_circuit(circuit)\n",
+    "print(circuit)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Use the `simplify_circuit` method to run a sequence of optimization passes on a circuit"
```
hay-k, line 332:
```suggestion
    "Use the `simplify_circuit` function to run a sequence of optimization passes on a circuit"
```

examples/usage.ipynb
```diff
@@ -260,6 +251,122 @@
...
+    "print('\\nState before the measurement:')\n",
+    "result = sim.simulate(circuit[:-1])\n",
+    "print(result)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Note that the above output vector represents the state before the measurement in the optimized circuit, not the original one, which would have the same phase for both terms. `simplify_circuit` has eliminated a `ZPowGate` which has no effect on the measurement."
```
hay-k, line 368: Let's remove this from here. I think the one in the Adonis section is enough. I didn't even check if it is true in this example as well, but the example is more complicated compared to the Adonis example and it is hard to calculate in your head what this comment says.

freetonik: Removed.

hay-k (approved):

PR #83, opened 2022-08-18
Update the demos

smite:
* Demos handle measurements more gracefully


examples/demo_common.py
```diff
@@ -29,36 +29,145 @@
 np.set_printoptions(precision=3)
 
 
+
+def get_measurements(circuit: cirq.Circuit) -> list[tuple[int, cirq.GateOperation]]:
```
freetonik, line 33:
```suggestion
np.set_printoptions(precision=3)


def get_measurements(circuit: cirq.Circuit) -> list[tuple[int, cirq.GateOperation]]:
```

ollityrkko (approved):

PR #88, opened 2022-09-26
Enable mypy support

freetonik:


src/cirq_iqm/devices/iqm_device.py
```diff
@@ -238,7 +238,7 @@ def route_circuit(
         new_measurements = []
         for _, op in measurement_ops:
             new_qubits = [final_qubit_mapping[q] for q in op.qubits]
-            new_measurement = cirq.measure(*new_qubits, key=op.gate.key)
+            new_measurement = cirq.measure(*new_qubits, key=op.gate.key)  # type: ignore
```
freetonik, line 241: Not sure there is a good way around this. The operations that end up in the `measurement_ops` are measurement operations, so they must have a measurement key. However, mypy cannot deduce this and complains about `key` not being found in `op.gate`.

kukushechkin: `findall_operations_with_gate_type` returns a tuple of three items where the last is operation gate casted to the requested gate type (`MeasurementGate`). That one should be recognized to have `.key`, but yeah, this will require modifying `measurement_ops` all over the place.

src/cirq_iqm/iqm_sampler.py
```diff
@@ -94,7 +94,8 @@ def run_sweep(
         mapped = program
         if self._qubit_mapping is not None:
             # apply the qubit_mapping
-            qubit_map = {cirq.NamedQubit(k): cirq.NamedQubit(v) for k, v in self._qubit_mapping.items()}
+            qubit_map: Union[dict['cirq.Qid', 'cirq.Qid'], Callable[['cirq.Qid'], 'cirq.Qid']] = \
```
kukushechkin, line 97: Does mypy recognize just `dict['cirq.Qid', 'cirq.Qid']` matching the `Union` function arguments?

freetonik: Good call, it does. Updated the file, thanks!

src/cirq_iqm/devices/iqm_device.py
```diff
@@ -92,7 +92,7 @@ def check_qubit_connectivity(cls, operation: cirq.Operation) -> None:
         """Raises a ValueError if operation acts on qubits that are not connected.
         """
         if len(operation.qubits) >= 2 and not isinstance(operation.gate, ops.MeasurementGate):
-            connection = set(cls.get_qubit_index(q) for q in operation.qubits)
+            connection = set(cls.get_qubit_index(q) for q in operation.qubits)  # type: ignore
```
q-mat-beu, line 95: Why not change the interface of `get_qubit_index`? This is the only place where it is used.

q-mat-beu (approved):

src/cirq_iqm/devices/iqm_device.py
freetonik: Thanks, that turned out to be a minor update. Check out the latest change please.

PR #92, opened 2022-10-27
Implement metadata for IQM devices

jkotilahti:
* Qubit connectivity and native gateset of a device are now defined using DeviceMetadata instead of class attributes, to have an interface consistent with other cirq* devices.
* IQMSampler will now by default use quantum architecture obtained from IQMClient to build the device metadata

### DIFF
--- a/docs/user_guide.rst
+++ b/docs/user_guide.rst
@@ -226,7 +226,7 @@ Note that the code snippet above assumes that you have set the variable ``iqm_se
 of the IQM server. By default, the latest calibration set is used for running the circuit. If you want to use
 a particular calibration set, provide the ``calibration_set_id`` argument. The sampler will by default use an
 :class:`.IQMDevice` created based on architecture data obtained from the server, which is then available in the
-:meth:`.IQMSampler.device` property. Alternatively, the device can be specified directly with the ``device`` argument.
+:attr:`.IQMSampler.device` property. Alternatively, the device can be specified directly with the ``device`` argument.
 
 If the IQM server you are connecting to requires authentication, you will also have to use
 `Cortex CLI <https://github.com/iqm-finland/cortex-cli>`_ to retrieve and automatically refresh access tokens,
--- a/docs/user_guide.rst
+++ b/docs/user_guide.rst
@@ -27,29 +27,32 @@ After installation Cirq on IQM can be imported in your Python code as follows:
 IQM's quantum devices
 ---------------------
 
-Cirq on IQM provides descriptions of IQM's quantum architectures as subclasses of :class:`.IQMDevice`.
-The abstract class :class:`.IQMDevice` itself is a subclass of :class:`cirq.devices.Device` and implements general
-functionality relevant to all IQM devices. Each device subclass describes the native gates and the connectivity of a particular architecture
-and the relevant functionality. As an example, let us import the class :class:`.Adonis`, which describes IQM's
-five-qubit architecture and view some of its properties contained in the class variables:
+Cirq on IQM provides descriptions of IQM's quantum architectures using the :class:`.IQMDevice` class, which is a
+subclass of :class:`cirq.devices.Device` and implements general functionality relevant to all IQM devices. The native
+gates and connectivity of the architecture are available in the :class:`.IQMDeviceMetadata` object returned by the
+:meth:`.IQMDevice.metadata` property of the device. It is possible to use :class:`.IQMDevice` class directly, but
+certain devices with predefined metadata are also available as subclasses of :class:`.IQMDevice`. As an example, let
+us import the class :class:`.Adonis`, which describes IQM's five-qubit architecture and view some of its properties
+contained in the :meth:`.Adonis.metadata` property:
 
 .. code-block:: python
 
     from cirq_iqm import Adonis
 
-    print(Adonis.QUBIT_COUNT)
-    print(Adonis.NATIVE_GATES)
-    print(Adonis.CONNECTIVITY)
+    adonis = Adonis()
+
+    print(adonis.metadata.qubit_set)
+    print(adonis.metadata.gateset)
+    print(adonis.metadata.nx_graph)
 
 
 IQM devices use :class:`cirq.NamedQubit` to represent their qubits. The names of the qubits consist of a prefix
 followed by a numeric index, so we have qubit names like ``QB1``, ``QB2``, etc. Note that we use 1-based
-indexing. The qubit connectivity information is stored using the qubit indices only. You can get the list of the qubits
-in a particular device by accessing the ``qubits`` attribute of a corresponding :class:`.IQMDevice` instance:
+indexing. You can get the list of the qubits in a particular device by accessing the ``qubits`` attribute of a
+corresponding :class:`.IQMDevice` instance:
 
 .. code-block:: python
 
-    adonis = Adonis()
     print(adonis.qubits)
 
 
@@ -214,14 +217,16 @@ instance and use its :meth:`~.IQMSampler.run` method to send a circuit for execu
 
    from cirq_iqm.iqm_sampler import IQMSampler
 
-   sampler = IQMSampler(iqm_server_url, adonis)
+   sampler = IQMSampler(iqm_server_url)
    result = sampler.run(circuit_1, repetitions=10)
    print(result.measurements['m'])
 
 
 Note that the code snippet above assumes that you have set the variable ``iqm_server_url`` to the URL
 of the IQM server. By default, the latest calibration set is used for running the circuit. If you want to use
-a particular calibration set, provide a ``calibration_set_id`` integer argument.
+a particular calibration set, provide a ``calibration_set_id`` integer argument. The sampler will by default use an
+:class:`.IQMDevice` created based on architecture data obtained from the server, which is then available in the
+:meth:`.IQMSampler.device` property. Alternatively, the device can be specified directly with the ``device`` argument.
 
 If the IQM server you are connecting to requires authentication, you will also have to use
 `Cortex CLI <https://github.com/iqm-finland/cortex-cli>`_ to retrieve and automatically refresh access tokens,
@@ -244,7 +249,7 @@ can be used to specify this correspondence.
 
     qubit_mapping = {'Alice': 'QB1', 'Bob': 'QB3'}
 
-    sampler = IQMSampler(iqm_server_url, adonis, qubit_mapping=qubit_mapping)
+    sampler = IQMSampler(iqm_server_url, qubit_mapping=qubit_mapping)
     result = sampler.run(decomposed_circuit_1, repetitions=10)
     print(result.measurements['m'])
 
--- a/docs/user_guide.rst
+++ b/docs/user_guide.rst
@@ -30,10 +30,10 @@ IQM's quantum devices
 Cirq on IQM provides descriptions of IQM's quantum architectures using the :class:`.IQMDevice` class, which is a
 subclass of :class:`cirq.devices.Device` and implements general functionality relevant to all IQM devices. The native
 gates and connectivity of the architecture are available in the :class:`.IQMDeviceMetadata` object returned by the
-:meth:`.IQMDevice.metadata` property of the device. It is possible to use :class:`.IQMDevice` class directly, but
-certain devices with predefined metadata are also available as subclasses of :class:`.IQMDevice`. As an example, let
-us import the class :class:`.Adonis`, which describes IQM's five-qubit architecture and view some of its properties
-contained in the :meth:`.Adonis.metadata` property:
+:attr:`.IQMDevice.metadata` property. It is possible to use the IQMDevice class directly, but
+certain devices with predefined metadata are also available as subclasses of IQMDevice. As an example, let
+us import the class :class:`.Adonis`, which describes IQM's five-qubit architecture, and view some of its
+properties contained in its ``metadata`` property:
 
 .. code-block:: python
 
@@ -48,7 +48,7 @@ contained in the :meth:`.Adonis.metadata` property:
 
 IQM devices use :class:`cirq.NamedQubit` to represent their qubits. The names of the qubits consist of a prefix
 followed by a numeric index, so we have qubit names like ``QB1``, ``QB2``, etc. Note that we use 1-based
-indexing. You can get the list of the qubits in a particular device by accessing the ``qubits`` attribute of a
+indexing. You can get the list of the qubits in a particular device by accessing the :attr:`qubits` attribute of a
 corresponding :class:`.IQMDevice` instance:
 
 .. code-block:: python
@@ -224,7 +224,7 @@ instance and use its :meth:`~.IQMSampler.run` method to send a circuit for execu
 
 Note that the code snippet above assumes that you have set the variable ``iqm_server_url`` to the URL
 of the IQM server. By default, the latest calibration set is used for running the circuit. If you want to use
-a particular calibration set, provide a ``calibration_set_id`` integer argument. The sampler will by default use an
+a particular calibration set, provide the ``calibration_set_id`` argument. The sampler will by default use an
 :class:`.IQMDevice` created based on architecture data obtained from the server, which is then available in the
 :meth:`.IQMSampler.device` property. Alternatively, the device can be specified directly with the ``device`` argument.
 
@@ -234,7 +234,7 @@ then set the ``IQM_TOKENS_FILE`` environment variable to use those tokens.
 See Cortex CLI's `documentation <https://iqm-finland.github.io/cortex-cli/readme.html>`_ for details.
 Alternatively, authorize with the IQM_AUTH_SERVER, IQM_AUTH_USERNAME and IQM_AUTH_PASSWORD environment variables
 or pass them as arguments to the constructor of :class:`.IQMProvider`, however this approach is less secure
-and considered as deprecated.
+and considered deprecated.
 
 When executing a circuit that uses something other than the device qubits, you need to either route it first
 as explained in :ref:`workflow 1 <workflow_1>` above,
--- a/src/cirq_iqm/devices/iqm_device.py
+++ b/src/cirq_iqm/devices/iqm_device.py
@@ -262,9 +262,5 @@ class IQMDevice(devices.Device):
 
         self.check_qubit_connectivity(operation)
 
-    @staticmethod
-    def _qubit_set_from_count(qubit_count: int) -> frozenset[cirq.NamedQubit]:
-        return frozenset(cirq.NamedQubit.range(1, qubit_count + 1, prefix=IQMDeviceMetadata.QUBIT_NAME_PREFIX))
-
     def __eq__(self, other):
         return self.__class__ == other.__class__ and self._metadata == other._metadata
--- a/src/cirq_iqm/devices/iqm_device.py
+++ b/src/cirq_iqm/devices/iqm_device.py
@@ -22,12 +22,14 @@ from __future__ import annotations
 
 import collections.abc as ca
 from math import pi as PI
-from typing import Optional, Union
+from typing import FrozenSet, Optional, Union
 import uuid
 
 import cirq
 from cirq import InsertStrategy, MeasurementGate, devices, ops, protocols
-from cirq.contrib.routing.router import nx, route_circuit
+from cirq.contrib.routing.router import route_circuit
+
+from .iqm_device_metadata import IQMDeviceMetadata
 
 
 def _verify_unique_measurement_keys(operations: ca.Iterable[cirq.Operation]) -> None:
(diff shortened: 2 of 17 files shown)


src/cirq_iqm/iqm_sampler.py
```diff
@@ -149,3 +166,13 @@ def _send_circuits(  # pylint: disable=too-many-arguments
...
+
+    def _metadata_from_architecture(self):
+        arch = self._client.get_quantum_architecture()
+        qubits = frozenset(cirq.NamedQubit(qb) for qb in arch.qubits)
+        connectivity = (
+            {int(qb.removeprefix(IQMDeviceMetadata.QUBIT_NAME_PREFIX)) for qb in edge}
+            for edge in arch.qubit_connectivity
+        )
+        gateset = cirq.Gateset(*(_IQM_CIRQ_OP_MAP[op] for op in arch.operations if _IQM_CIRQ_OP_MAP[op] is not None))
+        return IQMDeviceMetadata(qubits, connectivity, gateset)
```
q-mat-beu, line 178: I would prefer to have that as a method of `IQMDeviceMetadata` taking the `quantum_architecture` as an argument.

src/cirq_iqm/iqm_sampler.py
```diff
@@ -25,9 +25,17 @@
 from iqm_client.iqm_client import IQMClient
 import numpy as np
 
-from cirq_iqm import IQMDevice
+from cirq_iqm.devices.iqm_device import IQMDevice, IQMDeviceMetadata
 from cirq_iqm.iqm_operation_mapping import map_operation
 
+# Mapping from IQM operation names to cirq operations
+_IQM_CIRQ_OP_MAP = {
+    'phased_rx': cirq.ops.PhasedXPowGate,
```
q-mat-beu, line 33: There is the comment:
            "# XPow and YPow kept for convenience, Cirq does not know how to decompose them into PhasedX
            # so we would have to add those rules..."
If that one is still true we should map `phased_rx` to `PhasedXPowGate`, `XPowGate` and `YPowGate`.

src/cirq_iqm/iqm_sampler.py
jkotilahti: done

src/cirq_iqm/iqm_sampler.py
jkotilahti: Looks like that is still true, so I added `XPowGate` and `YPowGate` also here now.

docs/user_guide.rst
```diff
@@ -27,29 +27,32 @@ After installation Cirq on IQM can be imported in your Python code as follows:
...
 
-Cirq on IQM provides descriptions of IQM's quantum architectures as subclasses of :class:`.IQMDevice`.
-The abstract class :class:`.IQMDevice` itself is a subclass of :class:`cirq.devices.Device` and implements general
-functionality relevant to all IQM devices. Each device subclass describes the native gates and the connectivity of a particular architecture
-and the relevant functionality. As an example, let us import the class :class:`.Adonis`, which describes IQM's
-five-qubit architecture and view some of its properties contained in the class variables:
+Cirq on IQM provides descriptions of IQM's quantum architectures using the :class:`.IQMDevice` class, which is a
+subclass of :class:`cirq.devices.Device` and implements general functionality relevant to all IQM devices. The native
+gates and connectivity of the architecture are available in the :class:`.IQMDeviceMetadata` object returned by the
+:meth:`.IQMDevice.metadata` property of the device. It is possible to use :class:`.IQMDevice` class directly, but
```
smite, line 33:
```suggestion
:attr:`.IQMDevice.metadata` property of the device. It is possible to use :class:`.IQMDevice` class directly, but
```

docs/user_guide.rst
```diff
@@ -27,29 +27,32 @@ After installation Cirq on IQM can be imported in your Python code as follows:
...
+certain devices with predefined metadata are also available as subclasses of :class:`.IQMDevice`. As an example, let
+us import the class :class:`.Adonis`, which describes IQM's five-qubit architecture and view some of its properties
+contained in the :meth:`.Adonis.metadata` property:
```
smite, line 36:
```suggestion
contained in the :attr:`.Adonis.metadata` property:
```

docs/user_guide.rst
```diff
@@ -214,14 +217,16 @@ instance and use its :meth:`~.IQMSampler.run` method to send a circuit for execu
...
-   sampler = IQMSampler(iqm_server_url, adonis)
+   sampler = IQMSampler(iqm_server_url)
    result = sampler.run(circuit_1, repetitions=10)
    print(result.measurements['m'])
 
 
 Note that the code snippet above assumes that you have set the variable ``iqm_server_url`` to the URL
 of the IQM server. By default, the latest calibration set is used for running the circuit. If you want to use
-a particular calibration set, provide a ``calibration_set_id`` integer argument.
+a particular calibration set, provide a ``calibration_set_id`` integer argument. The sampler will by default use an
```
smite, line 227:
```suggestion
a particular calibration set, provide a ``calibration_set_id`` argument. The sampler will by default use an
```

src/cirq_iqm/devices/iqm_device.py
```diff
@@ -22,12 +22,14 @@
 
 import collections.abc as ca
 from math import pi as PI
-from typing import Optional, Union
+from typing import FrozenSet, Optional, Union
```
smite, line 25:
```suggestion
from typing import Optional, Union
```
Builtin types support the type annotation syntax directly, e.g. `frozenset[int]`

src/cirq_iqm/devices/iqm_device.py
```diff
@@ -269,5 +262,9 @@ def validate_operation(self, operation: cirq.Operation) -> None:
 
         self.check_qubit_connectivity(operation)
 
+    @staticmethod
+    def _qubit_set_from_count(qubit_count: int) -> FrozenSet[cirq.NamedQubit]:
```
smite, line 266:
```suggestion
    def _qubit_set_from_count(qubit_count: int) -> frozenset[cirq.NamedQubit]:
```

src/cirq_iqm/devices/iqm_device_metadata.py
```diff
@@ -0,0 +1,103 @@
...
+        connectivity: qubit connectivity graph of the device
+        gateset: Native gateset of the device. If None, a default IQM device gateset will be used.
+    """
+
+    QUBIT_NAME_PREFIX: str = 'QB'
+    """prefix for qubit names, to be followed by their numerical index"""
+
+    def __init__(  # pylint: disable=super-init-not-called
+        self,
+        qubits: FrozenSet[cirq.NamedQubit],
```
smite, line 50:
```suggestion
        qubits: frozenset[cirq.NamedQubit],
```

src/cirq_iqm/devices/iqm_device_metadata.py
```diff
@@ -0,0 +1,103 @@
...
+    'barrier': None,
+}
+
+
+@cirq.value.value_equality
+class IQMDeviceMetadata(devices.DeviceMetadata):
+    """Hardware metadata for IQM devices.
+
+    Args:
+        qubits: frozenset of qubits that exist on the device
```
smite, line 40:
```suggestion
        qubits: qubits that exist on the device
```
Types should not be repeated in the descriptions if possible, since they will be clear from the type annotations.

src/cirq_iqm/devices/iqm_device_metadata.py
```diff
@@ -0,0 +1,103 @@
...
+
+from typing import FrozenSet, Optional, Type, Union
+
+import cirq
+from cirq import NamedQubit, devices, ops
+from cirq.contrib.routing.router import nx
+from iqm_client import QuantumArchitectureSpecification
+
+# Mapping from IQM operation names to cirq operations
+_IQM_CIRQ_OP_MAP: dict[str, Optional[tuple[Union[Type[cirq.Gate], cirq.Gate, cirq.GateFamily], ...]]] = {
```
smite, line 25:
```suggestion
_IQM_CIRQ_OP_MAP: dict[str, Optional[tuple[Union[type[cirq.Gate], cirq.Gate, cirq.GateFamily], ...]]] = {
```

src/cirq_iqm/devices/iqm_device_metadata.py
```diff
@@ -0,0 +1,103 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""DeviceMetadata subtype for IQM devices."""
+from __future__ import annotations
+
+from typing import FrozenSet, Optional, Type, Union
```
smite, line 17:
```suggestion
from typing import Optional, Union
```

src/cirq_iqm/devices/iqm_device_metadata.py
```diff
@@ -0,0 +1,103 @@
...
+        gateset: Native gateset of the device. If None, a default IQM device gateset will be used.
+    """
+
+    QUBIT_NAME_PREFIX: str = 'QB'
+    """prefix for qubit names, to be followed by their numerical index"""
+
+    def __init__(  # pylint: disable=super-init-not-called
+        self,
+        qubits: FrozenSet[cirq.NamedQubit],
+        connectivity: tuple[set[int], ...],
```
smite, line 51:
```suggestion
        connectivity: tuple[set[cirq.NamedQubit], ...],
```

src/cirq_iqm/devices/iqm_device_metadata.py
```diff
@@ -0,0 +1,103 @@
...
+from iqm_client import QuantumArchitectureSpecification
+
+# Mapping from IQM operation names to cirq operations
+_IQM_CIRQ_OP_MAP: dict[str, Optional[tuple[Union[Type[cirq.Gate], cirq.Gate, cirq.GateFamily], ...]]] = {
+    # XPow and YPow kept for convenience, Cirq does not know how to decompose them into PhasedX
+    # so we would have to add those rules...
+    'phased_rx': (cirq.ops.PhasedXPowGate, cirq.ops.XPowGate, cirq.ops.YPowGate),
+    'cz': (cirq.ops.CZ,),
+    'measurement': (cirq.ops.MeasurementGate,),
+    'barrier': None,
```
smite, line 31:
```suggestion
    'barrier': (),
```

src/cirq_iqm/devices/iqm_device_metadata.py
```diff
@@ -0,0 +1,103 @@
...
+        connectivity = tuple(
+            {int(qb.removeprefix(IQMDeviceMetadata.QUBIT_NAME_PREFIX)) for qb in edge}
+            for edge in architecture.qubit_connectivity
+        )
+        gateset = cirq.Gateset(
+            *tuple(
+                cirq_op
+                for iqm_op in architecture.operations
+                if (cirq_ops := _IQM_CIRQ_OP_MAP[iqm_op]) is not None
+                for cirq_op in cirq_ops
```
smite, line 83:
```suggestion
            *(
                cirq_op
                for iqm_op in architecture.operations
                for cirq_op in _IQM_CIRQ_OP_MAP[iqm_op]
```

src/cirq_iqm/devices/iqm_device_metadata.py
```diff
@@ -0,0 +1,103 @@
...
+    QUBIT_NAME_PREFIX: str = 'QB'
+    """prefix for qubit names, to be followed by their numerical index"""
+
+    def __init__(  # pylint: disable=super-init-not-called
+        self,
+        qubits: FrozenSet[cirq.NamedQubit],
+        connectivity: tuple[set[int], ...],
+        gateset: Optional[cirq.Gateset] = None,
+    ):
+        """Construct an IQMDeviceMetadata object."""
```
smite, line 54: I would call `super().__init__(qubits, nx_graph)` somewhere here.

src/cirq_iqm/devices/iqm_device_metadata.py
```diff
@@ -0,0 +1,103 @@
...
+                cirq_op
+                for iqm_op in architecture.operations
+                if (cirq_ops := _IQM_CIRQ_OP_MAP[iqm_op]) is not None
+                for cirq_op in cirq_ops
+            )
+        )
+        return IQMDeviceMetadata(qubits, connectivity, gateset)
+
+    @property
+    def qubit_set(self) -> FrozenSet[cirq.NamedQubit]:
```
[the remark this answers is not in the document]
jkotilahti: Yes, but the reason I added it here is the more specific type hint than in the parent class (`NamedQubit` instead of `Qid`).

src/cirq_iqm/devices/iqm_device_metadata.py
jkotilahti: I had it previously, but removed it due to mypy complaining about the types. The super init uses `Qid` instead of `NamedQubit` for `self._qubits_set`, and that led to some type errors if we want to use NamedQubits.

smite: Hmm, NamedQubit inherits _BaseNamedQid, which inherits Qid, IMO it should work.

src/cirq_iqm/devices/iqm_device_metadata.py
smite: Type annotations of parameters should be general, the only thing we actually require of `connectivity` is that it can be iterated over (and same thing for the contents).

src/cirq_iqm/devices/iqm_device_metadata.py
jkotilahti: Yes, makes sense. Changed the arguments to use NamedQubits, implementation of the alternative initialization slightly different than suggested but I hope it's fine.

src/cirq_iqm/devices/iqm_device_metadata.py
```diff
@@ -0,0 +1,104 @@
...
+        else:
+            self._gateset = gateset
+
+    @classmethod
+    def from_architecture(cls, architecture: QuantumArchitectureSpecification) -> IQMDeviceMetadata:
+        """Returns device metadata object created based on architecture specification"""
+        qubits = frozenset(NamedQubit(qb) for qb in architecture.qubits)
+        connectivity = tuple({NamedQubit(qb) for qb in edge} for edge in architecture.qubit_connectivity)
+        gateset = cirq.Gateset(*(cirq_op for iqm_op in architecture.operations for cirq_op in _IQM_CIRQ_OP_MAP[iqm_op]))
+        return IQMDeviceMetadata(qubits, connectivity, gateset)
```
smite, line 76:
```suggestion
        return cls(qubits, connectivity, gateset)
```

src/cirq_iqm/devices/iqm_device_metadata.py
```diff
@@ -0,0 +1,104 @@
...
+
+    @classmethod
+    def from_qubit_indices(
+        cls, qubit_count: int, connectivity_indices: tuple[set[int], ...], gateset: Optional[cirq.Gateset] = None
+    ) -> IQMDeviceMetadata:
+        """Returns device metadata object created based on connectivity specified using qubit indices only."""
+        qubits = frozenset(NamedQubit.range(1, qubit_count + 1, prefix=IQMDeviceMetadata.QUBIT_NAME_PREFIX))
```
smite, line 83:
```suggestion
        qubits = frozenset(NamedQubit.range(1, qubit_count + 1, prefix=cls.QUBIT_NAME_PREFIX))
```

src/cirq_iqm/devices/iqm_device_metadata.py
```diff
@@ -0,0 +1,104 @@
...
+        connectivity = tuple(
+            {NamedQubit(f'{IQMDeviceMetadata.QUBIT_NAME_PREFIX}{qb}') for qb in edge} for edge in connectivity_indices
+        )
+        return IQMDeviceMetadata(qubits, connectivity, gateset)
```
smite, line 87:
```suggestion
        return cls(qubits, connectivity, gateset)
```

src/cirq_iqm/devices/iqm_device_metadata.py
```diff
@@ -0,0 +1,104 @@
...
+                ops.PhasedXPowGate, ops.XPowGate, ops.YPowGate, ops.MeasurementGate, ops.CZPowGate()
+            )
+        else:
+            self._gateset = gateset
+
+    @classmethod
+    def from_architecture(cls, architecture: QuantumArchitectureSpecification) -> IQMDeviceMetadata:
+        """Returns device metadata object created based on architecture specification"""
+        qubits = frozenset(NamedQubit(qb) for qb in architecture.qubits)
+        connectivity = tuple({NamedQubit(qb) for qb in edge} for edge in architecture.qubit_connectivity)
```
smite, line 74:
```suggestion
        qubits = tuple(NamedQubit(qb) for qb in architecture.qubits)
        connectivity = tuple(tuple(NamedQubit(qb) for qb in edge) for edge in architecture.qubit_connectivity)
```
Since they will be just iterated over once.

src/cirq_iqm/devices/iqm_device.py
```diff
@@ -51,48 +53,41 @@ class IQMDevice(devices.Device):
...
-    """native gate set of the device (gate families)"""
+    Args:
+        metadata: device metadata which contains the qubits, their connectivity, and the native gateset
+    """
 
-    NATIVE_GATE_INSTANCES: tuple[cirq.Gate, ...] = (ops.CZPowGate(),)
-    """native gate set of the device (individual gates)"""
+    def __init__(self, metadata: IQMDeviceMetadata):
+        self._metadata = metadata
+        self.qubits = tuple(sorted(self._metadata.qubit_set))
```
smite, line 63: The sorting here will probably sort `QB10` before `QB2`?

src/cirq_iqm/devices/iqm_device_metadata.py
smite: Looks good.

src/cirq_iqm/devices/iqm_device.py
smite: Could the metadata store a tuple of qubits instead of a set, and have the property convert it into a frozenset on demand? The tuple already has the correct order.

src/cirq_iqm/devices/iqm_device_metadata.py
jkotilahti: The problem is that when using `super().__init__` here, mypy thinks that `self._qubits_set` has `Qid`s instead of `NamedQubit`s (even though they really are named qubits). Then there is mypy error in `qubit_set()` since it should return `NamedQubit`. 

If we want use the super init but avoid those mypy errors it could be done by adding some `#type: ignore` comment or changing some return types to `Qid` instead of `NamedQubit`. Do you think that would be better?

src/cirq_iqm/devices/iqm_device.py
jkotilahti: >The sorting here will probably sort QB10 before QB2

No, `NamedQubits` with numbers are sorted correctly (https://quantumai.google/reference/python/cirq/NamedQubit: *NamedQubit has a lexicographic order. However, numbers within the name are handled correctly...*). I tested also by printing the `qubits` of Apollo. 

Since this is the case, I would not store them as tuple unless there is some other reason

jkotilahti:
CI pipeline is failing in `test_docs` due to this warning: 

```
Warning, treated as error:
failed to reach any of the inventories with the following issues:
intersphinx inventory 'https://docs.scipy.org/doc/scipy/objects.inv' not fetchable due to <class 'requests.exceptions.SSLError'>: HTTPSConnectionPool(host='docs.scipy.org', port=443): Max retries exceeded with url: /doc/scipy/objects.inv (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1129)')))
ERROR: InvocationError for command /home/runner/work/cirq-on-iqm/cirq-on-iqm/.tox/docs/bin/sphinx-build -W -b html docs /home/runner/work/cirq-on-iqm/cirq-on-iqm/build/sphinx/html (exited with code 2)
``` 

But locally `tox -e docs` works without warnings, so it looks like this is GitHub issue.

src/cirq_iqm/devices/iqm_device_metadata.py
jkotilahti: Changed IQMDeviceMetadata to use `Qid`s instead of `NamedQubit`s as discussed offline, which solves this and some of the other comments.

smite (approved):

PR #94, opened 2022-12-05
Bugfix: `initial_mapping` in route_circuit must be reversed. User guide updated.

smite:
* User guide updated.
* Reverse `initial_mapping` in route_circuit.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,17 @@
 Changelog
 =========
 
+Version 8.2
+===========
+
+* User guide updated.
+
+Bugfixes
+--------
+
+* :meth:`.IQMDevice.route_circuit` bugfix: ``initial_mapping`` must be reversed to match the
+  :mod:`cirq.contrib.routing.greedy` convention.
+
 Version 8.1
 ===========
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,7 +2,7 @@
 Changelog
 =========
 
-Version 8.2
+Version 9.0
 ===========
 
 * User guide updated.
--- a/docs/user_guide.rst
+++ b/docs/user_guide.rst
@@ -59,21 +59,17 @@ corresponding :class:`.IQMDevice` instance:
 Constructing circuits
 ---------------------
 
-There are two main workflows of using :class:`cirq.Circuit` instances with IQM devices:
+There are two main ways of constructing :class:`cirq.Circuit` instances for IQM devices:
 
-1. Create a ``Circuit`` instance, use arbitrary qubit names and types. There is no
-   validation of the operations when appending. At any point the user can apply :meth:`.IQMDevice.decompose_circuit`
-   to decompose the circuit contents into the native operation set of the device, or :meth:`.IQMDevice.route_circuit`
-   to route the circuit to the device connectivity and device qubits.
+1. Create a ``Circuit`` instance using arbitrary qubit names and types.
 2. Create a ``Circuit`` from an OpenQASM 2.0 program. The qubit names are determined by the OpenQASM ``qreg`` names,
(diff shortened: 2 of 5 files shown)


docs/user_guide.rst
```diff
@@ -96,42 +92,121 @@ This will result in the circuit
...
+Routing means transforming a circuit such that it acts on the device qubits, and respects the
+device connectivity.
 The method :meth:`.IQMDevice.route_circuit` accepts a :class:`cirq.Circuit` object as an argument,
 and returns the circuit routed against the device, acting on the device qubits instead of the
 arbitrary qubits we had originally.
 
 .. code-block:: python
 
-    routed_circuit_1 = adonis.route_circuit(decomposed_circuit_1)
+    routed_circuit_1 = adonis.route_circuit(decomposed_circuit)
```
hay-k, line 168:
```suggestion
    routed_circuit = adonis.route_circuit(decomposed_circuit)
```

docs/user_guide.rst
```diff
@@ -96,42 +92,121 @@ This will result in the circuit
...
     print(routed_circuit_1)
```
hay-k, line 169:
```suggestion
    routed_circuit = adonis.route_circuit(decomposed_circuit)
    print(routed_circuit)
```

src/cirq_iqm/devices/iqm_device.py
```diff
@@ -212,9 +213,16 @@ def route_circuit(
         for q in measurement_qubits:
             modified_circuit.append(cirq.I(q).with_tags(i_tag))
 
+        if initial_mapping is not None:
+            # reverse the mapping to match the cirq.contrib.routing.greedy convention
+            initial_mapping = {v: k for k, v in initial_mapping.items()}
```
q-mat-beu, line 218: Why don't we simply use the  cirq.contrib.routing.greedy convention?

freetonik: I suppose one reason is to allow to re-use the existing `qubit_mapping` values from users' old code?

CHANGELOG.rst
```diff
@@ -2,6 +2,17 @@
 Changelog
 =========
 
+Version 8.2
```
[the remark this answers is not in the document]
olliahonen: Should be major bump since this is a breaking change

src/cirq_iqm/devices/iqm_device.py
q-mat-beu: I think then we did not gain much from calling it `initial_mapping`. I would advocate to expose the cirq semantic to the user because the idea was to use cirq functionality instead of `qubit_mapping` to achieve qubit mapping.

smite: Otherwise I agree with @q-mat-beu, we should use Cirq conventions everywhere if possible.

q-mat-beu: Ok, convinced.

CHANGELOG.rst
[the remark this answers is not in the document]
smite: The docs have always described this behavior, so I'd consider it a bugfix.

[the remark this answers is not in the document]
olliahonen: Well, the docs were wrong 😄

freetonik (approved):

CHANGELOG.rst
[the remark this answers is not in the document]
olliahonen: ... and bugfixes _are_ breaking changes in the [ComVer scheme](https://gitlab.com/staltz/comver) we follow.

olliahonen (changes requested):

olliahonen (approved):

PR #112, opened 2023-05-31
implement heralding option

hay-k:


CHANGELOG.rst
```diff
@@ -2,6 +2,12 @@
 Changelog
 =========
 
+Version 11.9
+============
+
+* Add parameter ``heralding`` option to ``IQMSampler``. `#112 <https://github.com/iqm-finland/cirq-on-iqm/pull/112>`_
```
freetonik, line 8:
```suggestion
* Add ``heralding`` option to ``IQMSampler``. `#112 <https://github.com/iqm-finland/cirq-on-iqm/pull/112>`_
```

Just to be in line with [qiskit on iqm](https://github.com/iqm-finland/qiskit-on-iqm/pull/63/files#diff-2c623f3c6a917be56c59d43279244996836262cb1e12d9d0786c9c49eef6b43cR8)

hay-k: Well, in qiskit-iqm they are called option because Qiskit calls them options. Here they are just additional parameters to the initializer. Or maybe we can call it a property of the sampler class?

freetonik: Maybe then keep `parameter`, but drop `option` here? I simply (subjectively) find "parameter option" weird sounding :)

hay-k: ah, right. I didn't notice I have both word here. It is a typo indeed.

hay-k: Fixed!

q-mat-beu (approved):

PR #114, opened 2023-06-26
Add ability to abort submitted jobs

jkotilahti:
If the user interrupts the program with ctrl-c or similar (which creates a KeyboardInterrupt) when a job has been submitted to the server and the client is waiting for results, the submitted job is aborted.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,8 +2,8 @@
 Changelog
 =========
 
-Version 12.0
-============
+Version 11.11
+=============
 
 * Submitted job is aborted if the user interrupts the program while it is waiting for results. `#114 <https://github.com/iqm-finland/cirq-on-iqm/pull/114>`_
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,11 @@
 Changelog
 =========
 
+Version 12.0
+============
+
+* Submitted job is aborted if the user interrupts the program while it is waiting for results. `#114 <https://github.com/iqm-finland/cirq-on-iqm/pull/114>`_
+
 Version 11.10
 =============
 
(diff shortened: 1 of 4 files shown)


jkotilahti:
(CI will fail until https://github.com/iqm-finland/iqm-client/pull/89 is merged)

tests/test_iqm_sampler.py
```diff
@@ -216,6 +227,42 @@ def test_run_sweep_with_parameter_sweep(adonis_sampler, iqm_metadata, submit_cir
...
 
+@pytest.mark.usefixtures('unstub')
+def test_run_sweep_abort_job_successful(
+    adonis_sampler, circuit_physical, submit_circuits_default_kwargs, job_id, recwarn
+):
+    client = mock(IQMClient)
+    when(client).submit_circuits(ANY, **submit_circuits_default_kwargs).thenReturn(job_id)
+    when(client).wait_for_results(job_id).thenRaise(KeyboardInterrupt)
+    when(client).abort_job(job_id)
+    when(sys).exit().thenRaise(NotImplementedError)  # just for testing without actually exiting python
```
q-mat-beu, line 238: What about simply returning `None`?

CHANGELOG.rst
```diff
@@ -2,6 +2,11 @@
 Changelog
 =========
 
+Version 12.0
```
q-mat-beu, line 5: Similar to qiskit-on-iqm I don't think in this case we break any existing functionality.

tests/test_iqm_sampler.py
jkotilahti: The reason to use this instead of returning `None` is that otherwise there will be `'NoneType' object is not iterable` error in the last line of `client.run_sweep()` since `measurements` will be `None`.

CHANGELOG.rst
jkotilahti: Yes, I suppose these can be minor updates instead.

q-mat-beu (approved):

PR #121, opened 2023-12-22
Update project dependencies

rbrazinskas:
- Use latest version of iqm-client
- Switch to using cirq-core[contrib] instead of cirq to fix dependency conflict with pydantic version
- Update tools versions

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,9 +20,9 @@ classifiers = [
 requires-python = ">=3.9, <3.12"
 dependencies = [
     "numpy",
-    "cirq ~= 1.2",
+    "cirq-core[contrib] ~= 1.2",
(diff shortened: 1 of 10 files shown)


pyproject.toml
```diff
@@ -35,35 +35,29 @@ changelog = "https://github.com/iqm-finland/cirq-on-iqm/blob/main/CHANGELOG.rst"
...
-    "pytest-mypy == 0.9.1",
     "pytest-pylint == 0.21.0"
 ]
 cicd = [
     "twine ~= 4.0"
 ]
 
 
-[tool.setuptools.packages.find]
-where = ["src"]
```
smite, line 64: This is the default behavior I think

smite (approved):

PR #124, opened 2024-02-13
Upgrade to iqm-client 16.0

ollityrkko:


docs/user_guide.rst
```diff
@@ -279,13 +279,12 @@ The below table summarises the currently available options:
...
-     - Enable or disable server-side circuit duration checks. The default value is `True`, which means if any job is
-       estimated to take unreasonably long compared to the coherence times of the qubits, or too long in wall-clock
-       time, the server will reject it. This option can be used to disable this behaviour. In normal use, the
-       circuit duration check should always remain enabled.
+   * - `max_circuit_duration_over_t2`
+     - float
+     - 1.0
+     - Set server-side circuit disqualification threshold. If any job is estimated to take longer than the T2 time of the qubits
+       multiplied by this value the server will reject it. Setting this value to ``0.0`` will disable circuit duration check.
+       The default value ``None`` means the server default value will be used.
```
jkotilahti, line 287:
```suggestion
     - Set server-side circuit disqualification threshold. If any circuit in a job is estimated to take longer than the
       shortest T2 time of any qubit used in the circuit multiplied by this value, the server will reject the job.
       Setting this value to ``0.0`` will disable circuit duration check.
       The default value ``None`` means the server default value will be used.
```
Same as in qiskit-on-iqm.

src/iqm/cirq_iqm/iqm_sampler.py
```diff
@@ -57,8 +57,9 @@ class IQMSampler(cirq.work.Sampler):
             ID of the calibration set to use. If ``None``, use the latest one.
         run_sweep_timeout:
             timeout to poll sweep results in seconds.
-        circuit_duration_check: whether to enable or disable server-side circuit duration check
-        heralding_mode: Heralding mode to use during execution.
```
jkotilahti, line 61: Here also `heralding_mode` docstring got accidentally removed.

jkotilahti (changes requested):

src/iqm/cirq_iqm/iqm_sampler.py
ollityrkko: fixed

jkotilahti (approved):

PR #134, opened 2024-08-21
Allow inspecting a RunRequest without submitting it for execution

jkotilahti:
Requires https://github.com/iqm-finland/iqm-client/pull/129, tests will fail until it is merged.


Aerylia (commented):
lgtm

PR #136, opened 2024-09-04
Remove circuit validation when sending a circuit with IQMClient

Aerylia:
IQM client validates circuits before sending them to the quantum computer. This does not need to be done twice.

### DIFF
--- a/tests/test_iqm_sampler.py
+++ b/tests/test_iqm_sampler.py
@@ -16,7 +16,7 @@ import sys
 import uuid
 
 import cirq
-from mockito import ANY, mock, verify, when
+from mockito import ANY, expect, mock, verify, verifyNoUnwantedInteractions, when
 import numpy as np
 import pytest
 import sympy  # type: ignore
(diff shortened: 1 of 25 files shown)


tests/test_iqm_sampler.py
```diff
@@ -220,6 +224,8 @@ def test_run_sweep_has_heralding_mode_none_by_default(
     run_result = RunResult(status=Status.READY, measurements=[{'some stuff': [[0], [1]]}], metadata=iqm_metadata)
     kwargs = create_run_request_default_kwargs
     assert sampler._compiler_options.heralding_mode == HeraldingMode.NONE
+    kwargs = create_run_request_default_kwargs
```
kukushechkin, line 227: Either a diff artifact, or 2 identical lines added
Screenshot 2024-10-16 at 9 38 45

kukushechkin: The same on lines 249 and 257

tests/test_iqm_sampler.py
```diff
@@ -99,7 +100,10 @@ def run_request():
 
 @pytest.mark.usefixtures('unstub')
 def test_run_sweep_raises_with_non_physical_names(adonis_sampler, circuit_non_physical):
-    with pytest.raises(ValueError, match='Qubit not on device'):
+    when(adonis_sampler._client).get_quantum_architecture().thenReturn(
+        adonis_sampler._device.metadata.to_architecture()
+    )
+    with pytest.raises(CircuitExecutionError, match='Qubit Alice is not allowed as locus for measure'):
```
kukushechkin, line 106: I would add somewhere that this error comes from `iqm-client` and this is more of an integration test rather than `qiskit-iqm` unit test.

kukushechkin (approved):

PR #139, opened 2024-10-23
Use dynamic quantum architecture (DQA) for circuit decomposition/routing

jkotilahti:
- IQMDeviceMetadata can now be created from a DQA, and creating it from QuantumArchitectureSpecification (static architecture) is no longer supported.

- IQMDevice/IQMDeviceMetadata can still be created independently of the architecture/calset, so the hardcoded Adonis, Apollo etc. classes are not removed. They are currently used mainly for tests, and could be used for simulation.

- IQMSampler uses the specified calset (or server default calset if `calibration_set_id` not specified) to create the IQMDeviceMetadata/IQMDevice.

-  IQMSampler adds the calset id to the circuits it decomposes/routes. The calset id is used to warn the user of mismatches between the calset used for execution and the decomposed/routed circuits.

- Requires https://github.com/iqm-finland/iqm-client/pull/140 to be merged first, tests will fail until then. I have tested locally that all tests pass when used together with those changes.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,7 @@ Version 15.0
 
 * Use dynamic quantum architecture for circuit decomposition and routing. `#139 <https://github.com/iqm-finland/cirq-on-iqm/pull/139>`_
 * Require iqm-client >= 20.0. `#139 <https://github.com/iqm-finland/cirq-on-iqm/pull/139>`_
+* Disable attestations on ``gh-action-pypi-publish`` to fix failing PyPI publishing. `#139 <https://github.com/iqm-finland/cirq-on-iqm/pull/139>`_
 
 Version 14.6
 ============
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,12 @@
 Changelog
 =========
 
+Version 15.0
+============
+
+* Use dynamic quantum architecture for circuit decomposition and routing. `#139 <https://github.com/iqm-finland/cirq-on-iqm/pull/139>`_
+* Require iqm-client >= 20.0. `#139 <https://github.com/iqm-finland/cirq-on-iqm/pull/139>`_
+
 Version 14.6
 ============
 
--- a/src/iqm/cirq_iqm/devices/iqm_device_metadata.py
+++ b/src/iqm/cirq_iqm/devices/iqm_device_metadata.py
@@ -22,7 +22,7 @@ from cirq import Gate, NamedQid, devices, ops
 from cirq.contrib.routing.router import nx
 
 from iqm.cirq_iqm.iqm_operation_mapping import _IQM_CIRQ_OP_MAP
-from iqm.iqm_client import QuantumArchitectureSpecification
+from iqm.iqm_client import DynamicQuantumArchitecture
 
 
 @cirq.value.value_equality
@@ -35,6 +35,7 @@ class IQMDeviceMetadata(devices.DeviceMetadata):
         operations: Supported quantum operations of the device, mapping op types to their possible loci.
         gateset: Native gateset of the device. If None, a default IQM device gateset will be used.
         resonators: computational resonators of the device
+        architecture: architecture from which values of the other arguments were obtained
     """
 
     QUBIT_NAME_PREFIX: str = 'QB'
@@ -51,6 +52,7 @@ class IQMDeviceMetadata(devices.DeviceMetadata):
         operations: Optional[dict[type[cirq.Gate], list[tuple[cirq.NamedQid, ...]]]] = None,
         gateset: Optional[cirq.Gateset] = None,
         resonators: Iterable[NamedQid] = (),
+        architecture: Optional[DynamicQuantumArchitecture] = None,
     ):
         """Construct an IQMDeviceMetadata object."""
         nx_graph = nx.Graph()
@@ -79,6 +81,8 @@ class IQMDeviceMetadata(devices.DeviceMetadata):
             raise ValueError('Operations must be provided if a gateset is provided, it cannot be reconstructed.')
         self.operations = operations
 
+        self.architecture = architecture
+
     @property
     def resonator_set(self) -> FrozenSet[NamedQid]:
         """Returns the set of resonators on the device.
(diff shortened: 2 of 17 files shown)


src/iqm/cirq_iqm/iqm_sampler.py
```diff
@@ -65,12 +66,20 @@ def __init__(
...
             self._device = IQMDevice(device_metadata)
+            if calibration_set_id is None:
+                self._use_default_calibration_set = True
         else:
             self._device = device
-        self._calibration_set_id = calibration_set_id
+        if self._device.metadata.architecture is None:
+            self._calibration_set_id = calibration_set_id
+        else:
+            self._calibration_set_id = self._device.metadata.architecture.calibration_set_id
```
smite, line 82: I would suggest hiding `device` behind the `*` so the user has to be explicit if they want to provide it.

smite: What should happen if user gives both device and calibration_set_id?

src/iqm/cirq_iqm/iqm_sampler.py
```diff
@@ -148,6 +157,29 @@ def create_run_request(
...
+                    f'different than the current calibration set {self._calibration_set_id} of this sampler. '
+                    f'Decompose/route the circuits using this sampler to ensure successful execution.'
+                )
+            if self._use_default_calibration_set:
+                default_calset_id = self._client.get_dynamic_quantum_architecture(None).calibration_set_id
+                if self._calibration_set_id != default_calset_id:
+                    warnings.warn(
+                        f'Server default calibration set has changed from {self._calibration_set_id} '
+                        f'to {default_calset_id}. Use a new IQMSampler to decompose/route the circuits using '
+                        f'the new calibration set.'
```
smite, line 180:
```suggestion
                        'the new calibration set to ensure successful execution.'
```

src/iqm/cirq_iqm/devices/iqm_device_metadata.py
```diff
@@ -89,66 +94,55 @@ def resonator_set(self) -> FrozenSet[NamedQid]:
...
-        resonators = tuple(qb.name for qb in self.resonator_set)
-        connectivity = tuple(tuple(qb.name for qb in edge) for edge in self.nx_graph.edges())
-        operations: dict[str, list[tuple[str, ...]]] = {
-            iqm_op: [tuple(qb.name for qb in args) for args in qubits]
-            for cirq_op, qubits in self.operations.items()
-            for iqm_op, cirq_ops in _IQM_CIRQ_OP_MAP.items()
-            if cirq_op in cirq_ops
+            cirq_op: [get_qid_locus(locus) for locus in gate_info.loci]
+            for gate_name, gate_info in architecture.gates.items()
+            for cirq_op in _IQM_CIRQ_OP_MAP[gate_name]
```
smite, line 125: `operations` contains the full locus info for all gates, but it's not used for routing as far as I can tell.
`connectivity` is used (in the form of `IQMDevice._metadata.nx_graph`), but it does not represent any single 2q gate type (seems like the router assumes there is just one 2q gate).

Also, `connectivity` here will contain duplicate loci/edges. When it's converted into a graph the duplicates vanish.

jkotilahti: Right. As discussed, this problem already existed before this PR and is due to Cirq limitations. Probably can't solve it here. We can still look into a solution for it, but probably the PR should not be blocked from being merged because of this.

src/iqm/cirq_iqm/iqm_sampler.py
Aerylia: In that case, wouldn't you run into trouble if the device does not match the calibration_set_id information? Isn't it better to check and throw an error immediately?

src/iqm/cirq_iqm/devices/iqm_device_metadata.py
Aerylia: Does this connectivity work for Deneb-style devices? Where the router should not have the MOVE gate edges. If connectivity will turn into a an `nx_graph` does it makes sense to make labeled edges where the labels correspond to gates? Or is that something the router cannot work with either?

src/iqm/cirq_iqm/devices/iqm_device_metadata.py
```diff
@@ -35,6 +35,7 @@ class IQMDeviceMetadata(devices.DeviceMetadata):
         operations: Supported quantum operations of the device, mapping op types to their possible loci.
         gateset: Native gateset of the device. If None, a default IQM device gateset will be used.
         resonators: computational resonators of the device
+        architecture: architecture from which values of the other arguments were obtained
     """
 
     QUBIT_NAME_PREFIX: str = 'QB'
```
Aerylia, line 41: Is this still needed with DQA?

Aerylia (changes requested):

CHANGELOG.rst
```diff
@@ -2,6 +2,12 @@
 Changelog
 =========
 
+Version 15.0
+============
+
+* Use dynamic quantum architecture for circuit decomposition and routing. `#139 <https://github.com/iqm-finland/cirq-on-iqm/pull/139>`_
+* Require iqm-client >= 20.0. `#139 <https://github.com/iqm-finland/cirq-on-iqm/pull/139>`_
```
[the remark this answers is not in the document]
Aerylia: Oh, I just saw the PR. I somehow missed it.

src/iqm/cirq_iqm/devices/iqm_device_metadata.py
jkotilahti: It's still needed because it's used in `IQMDevice.get_qubit_index` and `IQMDeviceMetadata.from_qubit_indices`.

* `get_qubit_index` is not used anywhere in code. But could be useful for some purpose, not sure if there was some requirement to have it?

* `from_qubit_indices` makes creating the hardcoded devices like Apollo a bit nicer

Both are probably something we could get rid of if we really want, but I would still keep this and not change these things in this PR.

src/iqm/cirq_iqm/iqm_sampler.py
jkotilahti: True, if the given device is not compatible with the calset, submitting circuits would fail. I made now some changes to `IQMSampler.__init__`, so that now it checks device compatibility with the calset if `device` is explicitly passed.

Aerylia (approved):

src/iqm/cirq_iqm/devices/iqm_device_metadata.py
Aerylia: So the connectivity is used to build the `nx_graph` this is used in `route_circuit` by removing the resonators from it and then using Cirq's routing pass. After that, the MOVE gates are added. So there is no notion of gates in the `route_circuit` but the case of MOVE gates is handled through the resonator.

PR #145, opened 2025-02-10
switch order of route and decompose

manzanillo:
Currently, the example proposes to first run `decompose_circuit` followed by a `route_circuit`.
However, if the routing needs to introduce swap gates, these will stay in the circuit until execution where they lead to an error:
```
OperationNotSupportedError: <class 'cirq.ops.swap_gates.SwapPowGate'> not natively supported.
```

Therefore, the order of these two operations needs to be reversed.

### DIFF
--- a/examples/demo_iqm_execution.py
+++ b/examples/demo_iqm_execution.py
@@ -50,9 +50,9 @@ def demo_run_circuit() -> None:
 
     sampler = IQMSampler(os.environ['IQM_SERVER_URL'])
 
-    circuit_decomposed = sampler.device.decompose_circuit(circuit)
-    circuit_routed, _, _ = sampler.device.route_circuit(circuit_decomposed)
-    circuit = simplify_circuit(circuit_routed)
+    circuit_routed, _, _ = sampler.device.route_circuit(circuit)
+    circuit_decomposed = sampler.device.decompose_circuit(circuit_routed)
+    circuit = simplify_circuit(circuit_decomposed)
     print('\nTranspiled and routed circuit:\n')
     print(circuit)
     print('\n')


examples/demo_iqm_execution.py
```diff
@@ -50,9 +50,9 @@ def demo_run_circuit() -> None:
 
     sampler = IQMSampler(os.environ['IQM_SERVER_URL'])
 
-    circuit_decomposed = sampler.device.decompose_circuit(circuit)
-    circuit_routed, _, _ = sampler.device.route_circuit(circuit_decomposed)
-    circuit = simplify_circuit(circuit_routed)
+    circuit_routed, _, _ = sampler.device.route_circuit(circuit)
+    circuit_decomposed = sampler.device.decompose_circuit(circuit_routed)
+    circuit = simplify_circuit(circuit_decomposed)
```
smite, line 55: Did you test this against a station? Does `route_circuit` accept CX, which is not a native gate?

manzanillo: Yes, I tested it routing takes CNOT gates even though they are not native to `sampler.device`.

Aerylia: `route_circuit` uses a Cirq pass which is should be gateset agnostic.

As for Deneb, the `route_circuit` code checks if the device supports moves, if so, it will decompose after adding swaps and then calls `transpile_insert_moves_into_circuit`. 
See also https://github.com/iqm-finland/cirq-on-iqm/blob/main/src/iqm/cirq_iqm/devices/iqm_device.py#L256 

If you are supposed to decompose before routing, then I think `route_circuit` should return a decomposed circuit and thus call decompose by itself, returning decomposed swaps, and leave the docs as is. 
Otherwise, I think we should have a simple `transpile` wrapper that calls everything in the correct order depending on which device needs what, so that users don't run into these shenanigans.

smite: I think for now we can leave it as it is, if decomposition before routing is required it's up to the user to do it.

smite (approved):

PR #147, opened 2025-03-12
Expose the classically controlled PRX gate in Cirq

SalahedeenIssa:
In order to expose the classically controlled PRX gate in Cirq, a few things have to change: 

- The `cc_prx` operation in cirq does not give information directly on which qubit's measurement is acting as the classical control, only the key. As such, in order to fill out the "feedback_qubit" argument in cirq-iqm's version of the `cc_prx`, we have to take information from the preceding measurement instructions in `serialize.py`, rather than have it standalone like the other operations in `iqm_operation_mapping.py`. 
- The `cc_prx` operation in cirq does not have a gate associated with it, it is simply an operation. As such, when deserializing it, the current workflow for using gates in `_IQM_CIRQ_OP_MAP` won't support the `cc_prx` op.


SalahedeenIssa:
NOTE - I do not currently raise an error if the measurement condition occurs in a moment after the `cc_prx` operation's moment.

docs/user_guide.rst
```diff
@@ -259,6 +259,47 @@ optimizations. Let us try it out on our decomposed and routed circuit above:
     gates for a particular device. In the example above we don't have them, however it is generally a good idea to run
     decomposition once again after the simplification.
 
+Clasically Controlled Gates
+----------------------------
+Some IQM quantum computers support classically controlled gates, that is, gates that are executed
```
freetonik, line 264:
```suggestion
Classical control
-----------------
Some IQM quantum computers support classically-controlled gates, that is, gates that are executed
```

Just to follow the conventions of Cirq docs https://quantumai.google/cirq/build/classical_control

docs/user_guide.rst
```diff
@@ -259,6 +259,47 @@ optimizations. Let us try it out on our decomposed and routed circuit above:
...
+::
+
+    Alice: ─────────────────────────X───M───────────
+                                        ║
+    Bob: ───────────────────────────────╫───X───M───
+                                        ║   ║
+    Alices_Measurement_Register: ═══════@═══^═══════
+
+After an `X` gate is applied onto Alice, the measurement is stored
+in Alice's_Measurement_Register. If the result is 1 (which it should always be in this case),
```
freetonik, line 292:
```suggestion
in `Alices_Measurement_Register`. If the result is 1 (which it should always be in this case),
```

docs/user_guide.rst
```diff
@@ -259,6 +259,47 @@ optimizations. Let us try it out on our decomposed and routed circuit above:
...
+
+::
+
+    Alice: ─────────────────────────X───M───────────
+                                        ║
+    Bob: ───────────────────────────────╫───X───M───
+                                        ║   ║
+    Alices_Measurement_Register: ═══════@═══^═══════
+
+After an `X` gate is applied onto Alice, the measurement is stored
```
freetonik, line 291:
```suggestion
After an `X` gate is applied onto `Alice`, the measurement is stored
```

docs/user_guide.rst
```diff
@@ -259,6 +259,47 @@ optimizations. Let us try it out on our decomposed and routed circuit above:
...
+in Alice's_Measurement_Register. If the result is 1 (which it should always be in this case),
+another `X` gate is subsequently applied onto Bob. If it is zero, an identity gate of corresponding
+duration is applied instead.
+
+Executing the above circuit should always return `11`, even though Alice and Bob are never
```
freetonik, line 296:
```suggestion
Executing the above circuit should always return `11`, even though `Alice` and `Bob` are never
```

src/iqm/cirq_iqm/iqm_operation_mapping.py
```diff
@@ -123,4 +114,21 @@ def map_operation(operation: Operation) -> Instruction:
...
+            raise OperationNotSupportedError('Classically controlled prx gates can only have one condition')
+        if isinstance(operation._sub_operation.gate, (PhasedXPowGate, XPowGate, YPowGate)):
+            return Instruction(
+                name='cc_prx',
+                qubits=tuple(qubits),
+                args={
+                    'angle_t': operation._sub_operation.gate.exponent / 2,
+                    'phase_t': operation._sub_operation.gate.phase_exponent / 2,
+                    'feedback_qubit': 'n/a',
+                    'feedback_key': 'n/a',
```
freetonik, line 128: Is this the preferred way to specify the absence of those values, or should it be `None`?

SalahedeenIssa: I initially tried to set it to `None`, but I would get a type error before I could get the chance to change it

src/iqm/cirq_iqm/iqm_operation_mapping.py
```diff
@@ -123,4 +114,21 @@ def map_operation(operation: Operation) -> Instruction:
...
+    if isinstance(operation, cirq.ClassicallyControlledOperation):
+        if len(operation._conditions) > 1:
+            raise OperationNotSupportedError('Classically controlled prx gates can only have one condition')
+        if isinstance(operation._sub_operation.gate, (PhasedXPowGate, XPowGate, YPowGate)):
+            return Instruction(
+                name='cc_prx',
+                qubits=tuple(qubits),
+                args={
+                    'angle_t': operation._sub_operation.gate.exponent / 2,
+                    'phase_t': operation._sub_operation.gate.phase_exponent / 2,
```
Aerylia, line 126: Same here, the XpowerGate and YPowGate should not have a phase_exponent

tests/test_iqm_operation_mapping.py
```diff
@@ -114,3 +124,51 @@ def test_instruction_to_operation():
...
+
+
+def test_cc_prx_operation():
+    instruction = Instruction(
+        name='cc_prx',
+        qubits=('QB1',),
+        args={'angle_t': 0.5, 'phase_t': 0.25, 'feedback_qubit': 'COMP_R', 'feedback_key': 'test key'},
+    )
+    operation = instruction_to_operation(instruction)
+    assert isinstance(operation, ClassicallyControlledOperation)
```
Aerylia, line 136: You might want to check the angles and qubits here as well.

Aerylia (changes requested):

src/iqm/cirq_iqm/iqm_operation_mapping.py
```diff
@@ -78,24 +81,12 @@ def map_operation(operation: Operation) -> Instruction:
     """
     phased_rx_name = 'prx'
     qubits = [qubit.name if isinstance(qubit, NamedQid) else str(qubit) for qubit in operation.qubits]
-    if isinstance(operation.gate, PhasedXPowGate):
+    if isinstance(operation.gate, (PhasedXPowGate, XPowGate, YPowGate)):
```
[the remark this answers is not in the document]
SalahedeenIssa: I thought the same, but I think they do actually. I tried it locally and it worked fine, and on the cirq documentation of [XPowGate](https://quantumai.google/reference/python/cirq/XPowGate) and [YPowGate](https://quantumai.google/reference/python/cirq/YPowGate) they list `phase_exponent` as an attribute of each

Aerylia:
You will also need to write a changelog

src/iqm/cirq_iqm/iqm_operation_mapping.py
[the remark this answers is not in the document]
Aerylia: The hell? I checked the docs before writing this and was unable to find it <_<
If it works, it's fine.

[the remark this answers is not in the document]
SalahedeenIssa: Sounds good!

CHANGELOG.rst
```diff
@@ -2,6 +2,11 @@
 Changelog
 =========
 
+Version 16.1
+============
+
+* Expose the classically controlled prx gate `#147 <https://github.com/iqm-finland/cirq-on-iqm/pull/147>`_
```
Aerylia, line 8:
```suggestion
* Support for the classically controlled prx gate `#147 <https://github.com/iqm-finland/cirq-on-iqm/pull/147>`_
```

Aerylia (dismissed):
Minor suggestion, otherwise looks good.

smite:
> EDIT - Now I do, although it is based on order of total operations, not necessarily moment order. There is a small chance that a `cc_prx` operation occurs in the same moment as a measurement, which _could_ be acceptable, but is flagged as an error by my current workflow because the measurement operation technically is listed after the `cc_prx` .

Note that the Cirq Moment order does not determine the order in which the operations are scheduled in Cocos, only the partial order of the operations (the DAG order) matters to the scheduler.

docs/user_guide.rst
```diff
@@ -259,6 +259,49 @@ optimizations. Let us try it out on our decomposed and routed circuit above:
     gates for a particular device. In the example above we don't have them, however it is generally a good idea to run
     decomposition once again after the simplification.
 
+Classical control
+-----------------
+Some IQM quantum computers support classically-controlled gates, that is, gates that are executed
+conditionally depending on the result of a measurement preceding them in the quantum circuit. This
+support currently has several limitations:
+
+* Only the ``x``, ``y``, ``rx``, ``ry`` and ``r`` gates can be classically controlled.
```
smite, line 268: Cirq has no `r` gate, here it's called PhasedRXPowGate

smite (changes requested):

docs/user_guide.rst
smite: Also the other gate names should be fixed to match the Cirq conventions.

docs/user_guide.rst
```diff
@@ -259,6 +259,49 @@ optimizations. Let us try it out on our decomposed and routed circuit above:
     gates for a particular device. In the example above we don't have them, however it is generally a good idea to run
     decomposition once again after the simplification.
 
+Classical control
+-----------------
+Some IQM quantum computers support classically-controlled gates, that is, gates that are executed
```
smite, line 264:
```suggestion

Some IQM quantum computers support classically-controlled gates, that is, gates that are executed
```

docs/user_guide.rst
```diff
@@ -259,6 +259,49 @@ optimizations. Let us try it out on our decomposed and routed circuit above:
...
+conditionally depending on the result of a measurement preceding them in the quantum circuit. This
+support currently has several limitations:
+
+* Only the ``x``, ``y``, ``rx``, ``ry`` and ``r`` gates can be classically controlled.
+* The gates can only be conditioned on one classical bit, and the only control available is to
+  apply the gate if the bit is 1, and apply an identity gate if the bit is 0.
+* The availability of the controlled gates depends on the instrumentation of the quantum computer.
+
+The classical control can be applied on a circuit operation using :meth:`~cirq.Operation.with_classical_controls`:
```
smite, line 273:
```suggestion
The classical control can be applied on an operation using :meth:`~cirq.Operation.with_classical_controls`:
```
In Cirq CircuitOperation is a different class, so to avoid confusion

docs/user_guide.rst
```diff
@@ -259,6 +259,49 @@ optimizations. Let us try it out on our decomposed and routed circuit above:
...
+
+.. code-block:: python
+
+    import cirq
+    q1, q2 = cirq.NamedQubit('Alice'), cirq.NamedQubit('Bob')
+    circuit = cirq.Circuit(cirq.X(q1), cirq.measure(q1, key = 'Alices_Measurement_Register'),
+              cirq.X(q2).with_classical_controls('Alices_Measurement_Register'), cirq.measure(q2))
+    circuit
```
smite, line 281:
```suggestion
    q1, q2 = cirq.NamedQubit('Alice'), cirq.NamedQubit('Bob')
    circuit = cirq.Circuit(
        cirq.X(q1),
        cirq.measure(q1, key = 'A'),
        cirq.X(q2).with_classical_controls('A'),
        cirq.measure(q2),
    )
    print(circuit)
```
The measurement keys should not be called registers IMO, because the actual qubits in them can overlap. This is a more general system than in Qiskit.

docs/user_guide.rst
```diff
@@ -259,6 +259,49 @@ optimizations. Let us try it out on our decomposed and routed circuit above:
...
+
+Executing the above circuit should always return `11`, even though `Alice` and `Bob` are never
+actually entangled.
+
+.. note::
+
+   Because the gates can only take feedback from one classical bit you must place the measurement result
+   in a 1-bit classical register, ``Alices_Measurement_Register`` in the above example. The measurement
+   key used as the classical control must also be unique - running a classically controlled operation
+   where the condition is a key that is used to measure multiple qubits will result in an error.
```
smite, line 304:
```suggestion
   Because classically controlled gates can currently only take feedback from one classical bit you must condition
   them on a measurement key that only measures a single qubit.
```

src/iqm/cirq_iqm/devices/__init__.py
```diff
@@ -11,8 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""Definitions for varios IQM quantum devices for the Cirq framework.
-"""
+"""Definitions for varios IQM quantum devices for the Cirq framework."""
```
smite, line 14:
```suggestion
"""Definitions for various IQM quantum devices for the Cirq framework."""
```

src/iqm/cirq_iqm/iqm_operation_mapping.py
```diff
@@ -53,6 +63,8 @@ def instruction_to_operation(instr: Instruction) -> Operation:
         args['key'] = instr.args['key']
         args['num_qubits'] = len(instr.qubits)
     qubits = [NamedQid(qubit, dimension=2) for qubit in instr.qubits]
+    if instr.name == 'cc_prx':
```
smite, line 66: I can't refer to it direcly, but what is the "num_qubits" arg above?

smite: This function should be rewritten as an if/elif/else construct where for each operation type we look for the correct args.

src/iqm/cirq_iqm/iqm_operation_mapping.py
```diff
@@ -78,24 +90,12 @@ def map_operation(operation: Operation) -> Instruction:
     """
     phased_rx_name = 'prx'
```
smite, line 91: Can you remove this and refer to "prx" directly?

src/iqm/cirq_iqm/iqm_operation_mapping.py
```diff
@@ -123,4 +123,21 @@ def map_operation(operation: Operation) -> Instruction:
             args={},
         )
 
+    if isinstance(operation, ClassicallyControlledOperation):
+        if len(operation._conditions) > 1:
+            raise OperationNotSupportedError('Classically controlled prx gates can only have one condition')
```
smite, line 128:
```suggestion
            raise OperationNotSupportedError('Classically controlled gates can currently only have one condition.')
```

src/iqm/cirq_iqm/iqm_operation_mapping.py
```diff
@@ -123,4 +123,21 @@ def map_operation(operation: Operation) -> Instruction:
...
+                name='cc_prx',
+                qubits=tuple(qubits),
+                args={
+                    'angle_t': operation._sub_operation.gate.exponent / 2,
+                    'phase_t': operation._sub_operation.gate.phase_exponent / 2,
+                    'feedback_qubit': 'n/a',
+                    'feedback_key': 'n/a',
+                },
+            )
+
```
smite, line 140: Here we should raise an error that says we do not support classical control of this particular gate.

src/iqm/cirq_iqm/iqm_operation_mapping.py
```diff
@@ -123,4 +123,21 @@ def map_operation(operation: Operation) -> Instruction:
...
+            raise OperationNotSupportedError('Classically controlled prx gates can only have one condition')
+        if isinstance(operation._sub_operation.gate, (PhasedXPowGate, XPowGate, YPowGate)):
+            return Instruction(
+                name='cc_prx',
+                qubits=tuple(qubits),
+                args={
+                    'angle_t': operation._sub_operation.gate.exponent / 2,
+                    'phase_t': operation._sub_operation.gate.phase_exponent / 2,
+                    'feedback_qubit': 'n/a',
+                    'feedback_key': 'n/a',
```
smite, line 137: How about just empty strings?

src/iqm/cirq_iqm/serialize.py
```diff
@@ -30,8 +32,25 @@ def serialize_circuit(circuit: Circuit) -> iqm_client.Circuit:
     Returns:
         data transfer object representing the circuit
     """
-    instructions = tuple(map(map_operation, circuit.all_operations()))
-    return iqm_client.Circuit(name='Serialized from Cirq', instructions=instructions)
+    total_ops_list = [op for moment in circuit for op in moment]
```
smite, line 35:
```suggestion
    total_ops_list = list(circuit.all_operations())
```

src/iqm/cirq_iqm/serialize.py
```diff
@@ -30,8 +32,25 @@ def serialize_circuit(circuit: Circuit) -> iqm_client.Circuit:
...
+                    if m_idx > idx:
+                        raise OperationNotSupportedError("Measurement condition must precede cc_prx operation")
+                    if measurement_already_used:  # raise error if measurement has already been used as condition
+                        raise OperationNotSupportedError("Measurement condition for cc_prx must only be from one qubit")
+                    measurement_already_used = True  # change the flag to True now that measurement is used
+                    if len(i.qubits) > 1:
+                        raise OperationNotSupportedError("Measurement condition for cc_prx must only be from one qubit")
+                    feedback_qubit = i.qubits[0]
+                    instructions[idx].args["feedback_key"] = feedback_key
+                    instructions[idx].args["feedback_qubit"] = feedback_qubit
```
smite, line 52: I think you can avoid this double loop if you build a dict from measurement keys to the measurement ops, and a set of used keys.

smite: Then you can only iterate through the circuit once.

src/iqm/cirq_iqm/iqm_operation_mapping.py
SalahedeenIssa: I ensure that they can't use `sympy` expressions, but I think `cirq.Condition`s are fine. From the `cirq.ClassicallyControlledOperation` [page](https://quantumai.google/reference/python/cirq/ClassicallyControlledOperation): 

"Currently the only condition type is a measurement key. A measurement key evaluates to True iff any qubit in the corresponding measurement operation evaluated to a non-zero value."

Which I believe is in line with what we support

docs/user_guide.rst
```diff
@@ -259,6 +259,52 @@ optimizations. Let us try it out on our decomposed and routed circuit above:
...
+Classical control
+-----------------
+
+Some IQM quantum computers support classically-controlled gates, that is, gates that are executed
+conditionally depending on the result of a measurement preceding them in the quantum circuit. This
+support currently has several limitations:
+
+* Only the ``PhasedXPowGate``, ``XPowGate`` and ``YPowGate`` gates can be classically controlled.
+* The gates can only be conditioned on one classical bit, and the only control available is to
+  apply the gate if the bit is 1, and apply an identity gate if the bit is 0.
```
smite, line 271:
```suggestion
* The gates can only be conditioned on the measurement result of a single qubit, and the only control available
  is to apply the gate if the result is 1, and apply an identity gate if the result is 0.
```

docs/user_guide.rst
```diff
@@ -259,6 +259,52 @@ optimizations. Let us try it out on our decomposed and routed circuit above:
...
+    )
+    print(circuit)
+
+::
+
+    Alice: ─────────────────────────X───M───────────
+                                        ║
+    Bob: ───────────────────────────────╫───X───M───
+                                        ║   ║
+    Alices_Measurement_Register: ═══════@═══^═══════
```
smite, line 294: The output here needs to match the code.

docs/user_guide.rst
```diff
@@ -259,6 +259,52 @@ optimizations. Let us try it out on our decomposed and routed circuit above:
...
+
+After an `X` gate is applied on `Alice`, the qubit is measured and the result is stored
```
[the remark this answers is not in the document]
smite: also elsewhere.

src/iqm/cirq_iqm/serialize.py
SalahedeenIssa: I think it is easier now to use a for loop so that I can search for whether or not I need to include cc_prx support while building the `total_ops_list`

src/iqm/cirq_iqm/iqm_operation_mapping.py
```diff
@@ -123,4 +122,25 @@ def map_operation(operation: Operation) -> Instruction:
             args={},
         )
 
+    if isinstance(operation, ClassicallyControlledOperation):
+        if len(operation._conditions) > 1:
+            raise OperationNotSupportedError('Classically controlled gates can currently only have one condition.')
+        if isinstance(operation._conditions[0], (SympyCondition)):
+            raise OperationNotSupportedError('SympyConditions are not natively supported')
```
smite, line 129: I think both measurement keys as strings and MeasurementKey instances are converted into a `cirq.KeyCondition`, so it's safer to check for those, and raise an error if we see anything else.

smite:
```suggestion
        if not isinstance(operation._conditions[0], KeyCondition):
            raise OperationNotSupportedError('Only KeyConditions are supported in classically controlled operations.')
```

src/iqm/cirq_iqm/iqm_operation_mapping.py
```diff
@@ -123,4 +122,25 @@ def map_operation(operation: Operation) -> Instruction:
...
+                qubits=tuple(qubits),
+                args={
+                    'angle_t': operation._sub_operation.gate.exponent / 2,
+                    'phase_t': operation._sub_operation.gate.phase_exponent / 2,
+                    'feedback_qubit': '',
+                    'feedback_key': str(operation._conditions[0]),
+                },
+            )
+        raise OperationNotSupportedError(f'Classical control on the'
+                                         f' {type(operation._sub_operation.gate)} is not natively supported')
```
smite:
```suggestion
                                         f' {type(operation._sub_operation.gate)} is not supported.')
```

src/iqm/cirq_iqm/serialize.py
```diff
@@ -30,8 +31,35 @@ def serialize_circuit(circuit: Circuit) -> iqm_client.Circuit:
...
+    total_ops_list = []
+    cc_prx_support = False
+    for moment in circuit:
+        for op in moment:
+            if isinstance(op, ClassicallyControlledOperation):
+                cc_prx_support = True
+            total_ops_list.append(op)
+    instructions = list(map(map_operation, total_ops_list))
+
+    if cc_prx_support:
```
smite, line 43:
```suggestion
    circuit_ops = circuit.all_operations()
    instructions = list(map(map_operation, circuit_ops))

    if any(isinstance(op, ClassicallyControlledOperation) for op in circuit_ops):
```

SalahedeenIssa: Wouldn't this loop over the circuit more? Once to generate the circuit.all_operations, and once around the operations to locate the ClassicallyControlledOperation?

smite: Can you also merge iqm_operation_mapping.py into this file? No point in keeping them in separate modules IMO.

src/iqm/cirq_iqm/serialize.py
```diff
@@ -30,8 +31,35 @@ def serialize_circuit(circuit: Circuit) -> iqm_client.Circuit:
...
+                measurement_dict[inst.args["key"]] = (inst.qubits, idx)
+
+        for op, idx in cc_prx_dict.items():
+            m_idx = measurement_dict[op.args["feedback_key"]][1]
+            feedback_qubit = measurement_dict[op.args["feedback_key"]][0][0]
+            if len(feedback_qubit) > 1:
+                raise OperationNotSupportedError("Measurement condition for cc_prx must only be from one qubit")
+            if idx < m_idx:
+                raise OperationNotSupportedError("Measurement condition must precede cc_prx operation")
+            op.args["feedback_qubit"] = feedback_qubit[0]
```
smite:
```suggestion
        mkey_to_measurement = {}

        for inst in instructions:
            if inst.name == "measure":
                mkey_to_measurement[inst.args["key"]] = inst
            elif inst.name == "cc_prx":
                feedback_key = inst.args["feedback_key"]
                measurement = mkey_to_measurement.get(feedback_key)
                if measurement is None:
                    raise OperationNotSupportedError(f"cc_prx has feedback_key {feedback_key}, but no measure operation with that key precedes it.")
                if len(measurement.qubits) != 1:
                    raise OperationNotSupportedError("cc_prx must depend on the measurement result of a single qubit.")
                inst.args["feedback_qubit"] = measurement.qubits[0]
```

smite: If forgot
```

Comment (smite) [reply src/iqm/cirq_iqm/serialize.py]: So a measurement that is used as a feedback source needs both a key and a feedback_key for now, with the same value.

Comment (SalahedeenIssa) [reply src/iqm/cirq_iqm/serialize.py]: This is great, thanks!

Comment (smite) [reply src/iqm/cirq_iqm/serialize.py]: This isn't perfect yet, if you measure qubit A and use that as classical feedback on a prx on B, the measurement and prx might appear in any order when you iterate over the DAG. The user can force the measurement to come first using a barrier, but it's not an ideal user experience.

Comment (SalahedeenIssa) [reply src/iqm/cirq_iqm/serialize.py]: I think as long as the moment that contains the measurement is prior to the moment that contains the cc_prx it's okay (because I'm specifically looping over every moment in the circuit) The problem would arise I assume if they are in the same moment.

Comment (SalahedeenIssa) [reply src/iqm/cirq_iqm/serialize.py]: Sorry I'm not following here

> So a measurement that is used as a feedback source needs both a key and a feedback_key for now, with the same value.

File: docs/user_guide.rst [hunk 30]
```diff
@@ -261,47 +261,50 @@ optimizations. Let us try it out on our decomposed and routed circuit above:
...
+    Bob: ─────────╫───X───M────────────
+                  ║   ║
+    A: ═══════════@═══^════════════════
 
-After an `X` gate is applied onto `Alice`, the measurement is stored
-in `Alices_Measurement_Register`. If the result is 1 (which it should always be in this case),
-another `X` gate is subsequently applied onto `Bob`. If it is zero, an identity gate of corresponding
+After an ``X`` gate is applied on ``Alice``, the qubit is measured and the result is stored
+in the key `A`. If the result is one (which it should always be in this case),
+another `X` gate is subsequently applied on `Bob`. If it is zero, an identity gate of corresponding
```
Aerylia, line 298: Inconsistent quote delimitation. Unsure if this is on purpose.

src/iqm/cirq_iqm/serialize.py
```diff
@@ -32,25 +160,36 @@ def serialize_circuit(circuit: Circuit) -> iqm_client.Circuit:
...
-                    if len(i.qubits) > 1:
-                        raise OperationNotSupportedError("Measurement condition for cc_prx must only be from one qubit")
-                    feedback_qubit = i.qubits[0]
-                    instructions[idx].args["feedback_key"] = feedback_key
-                    instructions[idx].args["feedback_qubit"] = feedback_qubit
-    return iqm_client.Circuit(name="Serialized from Cirq", instructions=instructions)
+
+    if cc_prx_support:
+        mkey_to_measurement = {}
+        used_keys = set()
```
Aerylia, line 174: You shouldn't need `used_keys` as they are all in `mkey_to_measurement.keys()`

src/iqm/cirq_iqm/serialize.py
```diff
@@ -32,25 +160,36 @@ def serialize_circuit(circuit: Circuit) -> iqm_client.Circuit:
...
         data transfer object representing the circuit
     """
-    total_ops_list = [op for moment in circuit for op in moment]
+    total_ops_list = []
+    cc_prx_support = False
+    for moment in circuit:
+        for op in moment:
+            if isinstance(op, ClassicallyControlledOperation):
+                cc_prx_support = True
+            total_ops_list.append(op)
```
Aerylia, line 169: Do you need to go over the circuit to check if it has cc_prx? Can't you just assume `cc_prx_support=True` and run the code below?

src/iqm/cirq_iqm/serialize.py
```diff
@@ -16,11 +16,139 @@
...
+            args={'angle_t': operation.gate.exponent / 2, 'phase_t': operation.gate.phase_exponent / 2},
+        )
+    if isinstance(operation.gate, MeasurementGate):
+        if any(operation.gate.full_invert_mask()):
+            raise OperationNotSupportedError('Invert mask not supported')
+
+        return Instruction(
+            name='measure',
+            qubits=tuple(qubits),
+            args={'key': operation.gate.key},
```
Aerylia, line 109:
```suggestion
            args={'key': operation.gate.key, 'feedback_key': operation.gate.key},
```
As per @smite 's earlier comments. The 'key' is used for returning the measurement in the results and 'feedback_key' is used for the cc_prx.

src/iqm/cirq_iqm/serialize.py
Aerylia: The measurement that is used for the cc_prx also needs the `feedback_key` is what he meant.

src/iqm/cirq_iqm/serialize.py
```diff
@@ -30,7 +160,35 @@ def serialize_circuit(circuit: Circuit) -> iqm_client.Circuit:
...
+            elif inst.name == 'cc_prx':
+                feedback_key = inst.args['feedback_key']
+                measurement = mkey_to_measurement.get(feedback_key)
+                if measurement is None:
+                    raise OperationNotSupportedError(
+                        f'cc_prx has feedback_key {feedback_key}, but no measure operation with that key precedes it.'
+                    )
+                if len(measurement.qubits) != 1:
+                    raise OperationNotSupportedError('cc_prx must depend on the measurement result of a single qubit.')
+                inst.args['feedback_qubit'] = measurement.qubits[0]
```
Aerylia, line 190:
```suggestion
                inst.args['feedback_qubit'] = measurement.qubits[0]
                measurement['feedback_key'] = feedback_key
```
As per @smite 's earlier comment

src/iqm/cirq_iqm/serialize.py
Aerylia: I made a code proposal in a comment below

smite: A quantum circuit is effectively a DAG, (acyclic digraph), where the nodes are the gates and edges the qubit connections between them. The graph describes the (topological/causal) order in which the gates must be applied, but it's usually not a total order, only a partial order. This means there are multiple ways of iterating through the DAG in topological order, all equally valid. Cirq probably iterates in the order determined by the Moments, but that's only one way of doing it.

The problematic case is exactly if you have the measurement and the controlled PRX happen in the same Momement (on different qubits), in which case the Cirq iteration might produce them in the wrong order (prx first, then measurement), and we would falsely raise an error about an unknown measurement key.

src/iqm/cirq_iqm/serialize.py
```diff
@@ -15,10 +15,140 @@
...
+
+    gate_type = _IQM_CIRQ_OP_MAP[instr.name][0]
+    args_map = {'angle_t': 'exponent', 'phase_t': 'phase_exponent'}
+    args = {args_map[k]: v * 2 for k, v in instr.args.items() if k in args_map}
+    if 'key' in instr.args.keys():
+        args['key'] = instr.args['key']
+        args['num_qubits'] = len(instr.qubits)
+    qubits = [NamedQid(qubit, dimension=2) for qubit in instr.qubits]
+    if instr.name == 'cc_prx':
+        return gate_type(**args)(*qubits).with_classical_controls(instr.args['feedback_key'])
```
smite, line 71: Can you turn this into an `if inst.name == "aaa": ... elif inst.name == "bbb": ... else: ...` structure, so it'll be a bit easier to understand?

src/iqm/cirq_iqm/serialize.py
```diff
@@ -15,10 +15,140 @@
...
+            name='move',
+            qubits=tuple(qubits),
+            args={},
+        )
+
+    if isinstance(operation, ClassicallyControlledOperation):
+        if len(operation._conditions) > 1:
+            raise OperationNotSupportedError('Classically controlled prx gates can currently only have one condition.')
+        if not isinstance(operation._conditions[0], KeyCondition):
+            raise OperationNotSupportedError('Only KeyConditions are supported as classical controls')
```
smite, line 133:
```suggestion
            raise OperationNotSupportedError('Only KeyConditions are supported as classical controls.')
```

src/iqm/cirq_iqm/serialize.py
```diff
@@ -15,10 +15,140 @@
...
+    if isinstance(operation.gate, IQMMoveGate):
+        return Instruction(
+            name='move',
+            qubits=tuple(qubits),
+            args={},
+        )
+
+    if isinstance(operation, ClassicallyControlledOperation):
+        if len(operation._conditions) > 1:
+            raise OperationNotSupportedError('Classically controlled prx gates can currently only have one condition.')
```
smite, line 131:
```suggestion
            raise OperationNotSupportedError('Classically controlled gates can currently only have one condition.')
```

src/iqm/cirq_iqm/serialize.py
```diff
@@ -15,10 +15,140 @@
...
+                qubits=tuple(qubits),
+                args={
+                    'angle_t': operation._sub_operation.gate.exponent / 2,
+                    'phase_t': operation._sub_operation.gate.phase_exponent / 2,
+                    'feedback_qubit': '',
+                    'feedback_key': str(operation._conditions[0]),
+                },
+            )
+        raise OperationNotSupportedError(
+            f'Classical control on the {operation._sub_operation.gate} gate is not natively supported'
```
smite, line 146:
```suggestion
            f'Classical control on the {operation._sub_operation.gate} gate is not supported'
```

src/iqm/cirq_iqm/serialize.py
smite: You could just loop once over `circuit.all_operations()`, and convert the Cirq ops one by one to IQM ops, and then run the rest of the loop code below.

Aerylia: Yeah, but if ccprx is found, you will loop over the circuit regardless. So I guess this is a speedup for circuits without ccprxs

src/iqm/cirq_iqm/serialize.py
SalahedeenIssa: And PyCharm is not happy with me running `if instr.args.keys() = bla` followed by `elif inst.name = bla`

SalahedeenIssa: Nevermind, it turns out that only `measure` instructions have `key` in `instr.args[`key`], and that they are always required to have a key I believe. So `if 'key' in instr.args.keys()` and `if instr.name == "measure"` are effectively the same, and I've changed it to that

src/iqm/cirq_iqm/iqm_operation_mapping.py
Aerylia: I think 'n/a', 'tmp', or 'to be filled in' are sensible strings to put in. The circuit should never contain these strings, unless something went wrong in the serialization. If that happens and the string is empty, it might take a while to realize that it is a problem in serialization.

SalahedeenIssa: That's a fair point too. Should I change it back or wait for @smite to respond?

src/iqm/cirq_iqm/serialize.py
Aerylia: This difference affects performance. Not sure which is better, but I don't think it matters.

src/iqm/cirq_iqm/serialize.py
Aerylia: This has been done.

src/iqm/cirq_iqm/iqm_operation_mapping.py
Aerylia: Maybe check tomorrow during daily. I don't know how fast this needs to be merged.

SalahedeenIssa: I don't know how urgent it is to merge this, but I'm also going to be missing the next two weeks starting tomorrow, so I wanted to have this done before my vacation. I don't mind waiting though

SalahedeenIssa: I feel like this particular issue isn't a blocker though if it comes down to it, we can always go in and change it later

perjlieb (approved):
🎏

